Issue #69, opened 2020-08-20
[FR] Support specifying minimum TLS version for Azure MySQL


claassen:
**Is your feature request related to a problem? Please describe.**
No default TLS minimum version is configured when use_tls is set to true.

**Describe the solution you'd like**
Would like to be able to specify the minimum TLS version when creating Azure MySQL database.

**Describe alternatives you've considered**
N/A

**Priority**
High

**Priority Context**
Security requirement

**Platform** 
Azure

**Applicable Services**
MySQL

**Proposed Solution**
Use https://www.terraform.io/docs/providers/azurerm/r/mysql_server.html#ssl_minimal_tls_version_enforced

omerbensaadon:
spoke with @claassen on Slack, he will submit a PR to fix this (thank you!)

PR #71, opened 2020-08-21
Support for specifying minimum TLS version in Azure MySQL and Azure Redis

claassen:
Implements https://github.com/pivotal/cloud-service-broker/issues/69


omerbensaadon:
Fixes #69

PR #88, opened 2020-09-14
added spring-boot app for gcpbigquery,gcpspanner,gcpbucket

svennela:


svennela:
I am going update testscript with app push

PR #98, opened 2020-09-23
Fix typo

mogul:


omerbensaadon (approved):
good eye, thanks for looking out!

PR #104, opened 2020-09-25
Accurately describe the functionality

mogul:


mogul:
Actually I'm not sure this is right... I assumed "AWS Service Broker" referred to the AWS brokerpak, but maybe you're referring instead to the AWS-provided service broker? In any case, some clarification is needed here.

omerbensaadon:
added clarity here, it is the AWS-provided service broker

omerbensaadon (approved):

Issue #77, opened 2020-09-03
[FR] Support AWS IAM Role Profile


xyloman:
**Is your feature request related to a problem? Please describe.**

We are attempting to remove the use of access keys and secrets from all AWS Service interactions.  In doing so we would require the use of an IAM Role Profile to be leveraged when using the service broker instead of AWS Access Key and Secret.  This might require the broker to run on a VM instead of an application on cf when running on cf-for-vms.  However, when running on cf-for-k8s EKS does support the ability to leverage a [k8s service accounts](https://aws.amazon.com/about-aws/whats-new/2019/09/amazon-eks-adds-support-to-assign-iam-permissions-to-kubernetes-service-accounts/#:~:text=Now%2C%20Amazon%20EKS%20allows%20you,individual%20pod%20that%20you%20run.

**Describe the solution you'd like**
Allow for the configuration of IAM Role Profile on an EC2 instance via Instance Profile.

**Describe alternatives you've considered**
Continue to use AWS Access Key and Secrets and work with our security team to get an exception.

**Additional Context**
This has come up as apart of our [HITRUST](https://hitrustalliance.net/) compliance audit.  Still waiting to hear if its an absolute necessity or a delighter.

**Priority**
Medium unless it causes us to fail the audit then High.

**Priority Context**
Yes this would ensure that the platform continues to use instance and role profiles for obtaining temporary use access keys and secrets which would prevent the possibility of a developer or operator leaking them.

**Platform** 
AWS

**Applicable Services**
Provisioning of AWS Managed Services would be interesting to see how this would work for brokered services like Dynamo or AWS ElasticSearch Service which could use AWS IAM credentials instead of basic auth for authentication.  I believe that would be scoped to a different issue.  This should strictly be on the provisioning of resources.  We have teams that have done similarly with terraform recently when running on a VM I believe.

omerbensaadon:
Hey Bryan! Will bring this up at standup... may require some further questions/a spike... 

Initial take on this is that we probably won't prioritize this until after GA (EOM Sep)

Logiraptor:
Just to add more context: This is a hard requirement generally for many security-conscious organizations like Banks and Government. Additionally, in some government AWS regions it is impossible to use an Access Key and Secret for longer than 1 hour, so using an IAM instance profile is the preferred option for long-lived credentials.

Logiraptor:
I can see three potential options for implementing this on cf-for-vms:

[...]

PR #122, opened 2020-11-04
Update required permissions in the AWS installation guide

Logiraptor:
Hello, I found the permissions were incorrect while working on #77. This PR updates the list to be at least correct for creating RDS instances. I have not verified the other AWS services, so there could still be missing permissions for those.


omerbensaadon (commented):
Thanks so much for your contribution! Sincerely appreciate it you saving us the trouble. Will have ENG approve so they can review

omerbensaadon (approved):
Thank you!!

PR #142, opened 2020-11-25
AWS Brokerpak Clean Up

michaelmccaskill:
### Changes

- Major refactor of layout of AWS services by extracting provisioning/binding into separate Terraform folders with separate files for data, outputs, etc. as this is considered Terraform best practices
- Added ability to specify RDS subnet group. This is useful if your VPC already has an existing subnet group that you want to use


erniebilling (approved):
Will be re-adding the subsume support for postgres, but looks good.

PR #168, opened 2021-01-21
Enlarge DB column to allow for larger Terraform state

mogul:
Expands the size of the Workspace column to handle deployments where the Terraform workspace is greater than 64K. 

We need this for our eks-brokerpak since the `.tfstate` ends up about 196K...!

Mediumtext allows for workspaces up to 16384K.

### DIFF
--- a/db_service/migrations.go
+++ b/db_service/migrations.go
@@ -22,7 +22,7 @@ import (
 	"github.com/jinzhu/gorm"
 )
 
-const numMigrations = 7
+const numMigrations = 8
 
 // runs schema migrations on the provided service broker database to get it up to date
 func RunMigrations(db *gorm.DB) error {
@@ -78,6 +78,10 @@ func RunMigrations(db *gorm.DB) error {
 		} else {
 			return db.Model(&models.ProvisionRequestDetailsV2{}).ModifyColumn("request_details", "text").Error
 		}
+	}
+
+	migrations[7] = func() error { // v4.2.0
+		return autoMigrateTables(db, &models.TerraformDeploymentV2{})
 	}
 
 	var lastMigrationNumber = -1
--- a/db_service/models/db.go
+++ b/db_service/models/db.go
@@ -72,4 +72,4 @@ type CloudOperation CloudOperationV1
 
 // TerraformDeployment holds Terraform state and plan information for resources
 // that use that execution system.
-type TerraformDeployment TerraformDeploymentV1
+type TerraformDeployment TerraformDeploymentV2
--- a/db_service/models/historical_db.go
+++ b/db_service/models/historical_db.go
@@ -225,6 +225,29 @@ type TerraformDeploymentV1 struct {
 	LastOperationMessage string `sql:"type:text"`
 }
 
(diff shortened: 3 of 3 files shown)


erniebilling (approved):

mogul:
@erniebilling This was my first time working with Go DB facilities and I'm confused about what I'm seeing in the git history, and observing in my deployment. Referring to [this earlier commit over here](https://github.com/cloudfoundry-incubator/cloud-service-broker/blame/master/db_service/models/historical_db.go#L215) (I had only just rebased and didn't see that commit in the history right before I submitted the PR): I was under the impression that we couldn't just update the type of the `Workspace` string in `TerraformDeploymentV1`, because that wouldn't perform any migrations for the column type in existing deployments. That's why I added TerraformDeploymentV2 and the other changes to perform the migration, rather than making a commit like that earlier one. 

It looks like my PR may correct an oversight and inadvertently save other folks from having a mismatched column type left in their deployment. As evidence, in my long-ago-provisioned, pre-existing DB, I see that the type of the `workspace` column is still `text` even after my PR was merged:

```
mysql> describe terraform_deployments;
+------------------------+--------------+------+-----+---------+-------+
| Field                  | Type         | Null | Key | Default | Extra |
+------------------------+--------------+------+-----+---------+-------+
| id                     | varchar(255) | NO   | PRI |         |       |
| created_at             | timestamp    | YES  |     | NULL    |       |
| updated_at             | timestamp    | YES  |     | NULL    |       |
| deleted_at             | timestamp    | YES  |     | NULL    |       |
| workspace              | text         | YES  |     | NULL    |       |
| last_operation_type    | varchar(255) | YES  |     | NULL    |       |
| last_operation_state   | varchar(255) | YES  |     | NULL    |       |
| last_operation_message | text         | YES  |     | NULL    |       |
+------------------------+--------------+------+-----+---------+-------+
8 rows in set (0.03 sec)
```

On this same point: I expected to find the broker using the table corresponding to `TerraformDeploymentV2` (`terraform_deployment_v2`) after my PR (as of 0.2.2), but I'm not seeing that... The broker is still stashing rows in the older `terraform_deployments` table (and for that matter, the old table still exists; not sure if it should). Is there something else I needed to do to get the broker using the new table? Was there some other element of doing migrations in Go that I missed?

(Incidentally, my PR worked with a large Terraform workspace locally because the DB used for testing was provisioned *after* that other commit, and therefore *always* had a `Workspace` type of `mediumtext` while I was testing.)

mogul:
OK, I figured this out by retracing @basiszwo's steps in these two commits: 😳 🤦  
https://github.com/GSA/cloud-service-broker/commit/bf2000b1e0a213686d86840a1f8a925d36cefc8c#diff-a2735d28ece586878d8d42a56a13141f85f163caafbbe681307ad42910107b74R138-R143
https://github.com/GSA/cloud-service-broker/commit/bd36dcb075b8b758d1a0b7021d746c8c983d25a9#diff-0ace05affb10ac2c4aa1edf4026bf9020a7ce7e09fbf1e1e6bfa8a7438e4f07eR74-R82

New PR that addresses this problem incoming...!

PR #200, opened 2021-03-12
chore: log concurrent ID when running examples

blgm:
- this makes it easier to work out which log line relates to which
example
- also logs a summary at the end

[#177317568](https://www.pivotaltracker.com/story/show/177317568)


pivotal-marcela-campo (approved):
Nice improvements!

PR #221, opened 2021-04-12
chore: broker logs should show correlation ID

blgm:
log lines relating to a request should show the correlation ID so that
they can be differentiated when there are concurrent requests

[#177289977](https://www.pivotaltracker.com/story/show/177289977)


FelisiaM (approved):
👍

FelisiaM:
I think the code looks right, but my only concern is that with CF as the platform there is no `correlationId` or `requestIdentity` header sent. So even though this will attempt to log the actual change should happen in SAPI.

blgm:
The `brokerapi` modules guarantees that there is at least a correlation ID even when none is sent by the platform. This means that it will be possible to distinguish between requests when concurrent requests come in.

PR #228, opened 2021-04-16
Removes the unused policy actuator.

josephlewis42:
Policy actuation was a leftover of the GCP Service Broker that allowed administrators to define policy rules for their developers that modified or restricted values. This isn't necessary in a Terraform based world where they can just modify the plans instead.


blgm (approved):

blgm:
Thank you @josephlewis42!

Issue #99, opened 2020-09-23
[BUG] "Failed to query available provider packages" despite terraform packages being available


mogul:
My brokerpak attempts to use the Terraform registry despite provider packages being available locally.

## Description
GIVEN the file `manifest.yml` contains (note the version of Terraform):
```
packversion: 1
name: solr-services-pak
version: 1.0.0
metadata:
  author: [redacted]
platforms:
- os: linux
  arch: "386"
- os: linux
  arch: amd64
terraform_binaries:
- name: terraform
  version: 0.13.3
  source: https://github.com/hashicorp/terraform/archive/v0.13.3.zip  
- name: terraform-provider-random
  version: 2.3.0
  source: https://releases.hashicorp.com/terraform-provider-random/2.3.0/terraform-provider-random_2.3.0_linux_amd64.zip
- name: terraform-provider-kubernetes
  version: 1.13.2
  source: https://releases.hashicorp.com/terraform-provider-kubernetes/1.13.2/terraform-provider-kubernetes_1.13.2_linux_amd64.zip
- name: terraform-provider-helm
  version: 1.3.0
  source: https://releases.hashicorp.com/terraform-provider-helm/1.3.0/terraform-provider-helm_1.3.0_linux_amd64.zip
service_definitions:
- services/solr-operator.yml
parameters: []
required_env_variables: []
env_config_mapping: {}
```
...AND the file `services/solr-operator.yml` contains:
```
version: 1
name: solr-operator
id: f145c5aa-4cee-4570-8a95-9a65f0d8d9da
description: Fault-tolerant and highly-available distributed indexing and searching using Apache Solr, in the Kubernetes cluster of your choice
display_name: Apache SolrCloud
image_url: https://lucene.apache.org/theme/images/solr/identity/Solr_Logo_on_white.png
documentation_url: https://lucene.apache.org/solr/resources.html
support_url: https://github.com/GSA/solr-brokerpak
tags: [apache, search, index, k8s]
plans:
- name: base
  id: 1779d7d5-874a-4352-b9c4-877be1f0745b
  description: Establish a beachhead in the provided k8s where you want SolrCloud services available
  display_name: SolrCloud beachhead in k8s
  bullets:
  - "REQUIRED prerequisite: Create an operator before creating instances of the SolrCloud service"
  properties: {}
provision:
  plan_inputs: []
  user_inputs:
  - required: true
    field_name: cluster_id
    type: string
[...]

PR #233, opened 2021-04-26
Support for terraform 0.15 (and lower)

dmachard:
Hi,  

This is a proposal to support terraform > 0.12 #99 . Below some explanation.

- With this PR, it's not longer necessary to declare terraform binaries in the manifest yaml. Just the binary of terraform must be declared.

```
platforms:
- os: linux
  arch: amd64
terraform_binaries:
- name: terraform
  version: 0.15.0
  source: https://releases.hashicorp.com/terraform/0.15.0/terraform_0.15.0_linux_amd64.zip
```

- During the build of the brokerpak, terraform providers to be integrated are automatically detected and downloaded.
Each service is inspected to get terraform files and get providers used.

```
2021/04/26 19:29:50 Packing terraform providers...
2021/04/26 19:29:54     terraform-provider-null_v3.1.0_x5
```

Providers binaries are saved in the same way of the terraform init command. In the folder "plugins" for terraform 0.12 or in the folder "providers" for terraform > 0.12.

- Finally plugins are loaded as usual with the -plugin-dir option.

With this approch, terraform 0.12 and greater are supported.

Denis

### DIFF
--- a/pkg/brokerpak/manifest.go
+++ b/pkg/brokerpak/manifest.go
@@ -16,11 +16,14 @@
 package brokerpak
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"github.com/cloudfoundry-incubator/cloud-service-broker/pkg/providers/tf"
 	"github.com/cloudfoundry-incubator/cloud-service-broker/pkg/validation"
@@ -132,6 +135,11 @@ func (m *Manifest) Pack(base, dest string) error {
 		return err
 	}
 
+	log.Println("Packing terraform providers...")
+	if err := m.packProviders(dir, base); err != nil {
+		return err
+	}
+
 	log.Println("Creating archive:", dest)
 	return ziputil.Archive(dir, dest)
 }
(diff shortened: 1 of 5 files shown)


blgm:
Hi @dmachard. Thank you for this PR. I am completely in agreement with you that the CSB must eventually support Terraform versions higher than v1.12.

I think this PR introduces an alternative syntax for the `manifest.yaml` in the brokerpak. The old syntax (specifying the providers explicitly) still works, and there is a new syntax where the providers do not have to be specified explicitly. This supports Terraform v0.15, and is a more elegant solution.

I do have a concern that having more than one syntax introduces complexity that makes future development harder. For instance, if a future contributor only tests their changes with the old syntax, would they know whether or not they were breaking the new syntax? It may be better to allow only a single syntax so that future development does not introduce unexpected breaks due to testing gaps.

I'm thinking that the best way forward would be that at some point in the future we could merge this PR and also introduce some deprecation warnings for the old syntax, giving users a small number of releases during which to update their `mainifest.yaml` files. The old syntax could then be completely removed. I can also imagine a future where a brokerpak may have more than one Terraform version. This would be to support the migration of existing service instances to a newer Terraform version. There's an argument that if we want to change the syntax, we should make it so that more than one Terraform version could be supported in the future. Although there's also an argument that this would be over-engineering the solution. 

At the moment the team maintaining the CSB is building up context as previous maintainers have moved on. We would prefer to wait until a time when we have more context on the CSB so that we understand all the implications of the code changes, and can be sure to manage the syntax change successfully. Therefore I propose that this PR should not be merged just yet, but should be kept open so that it can be reviewed again in the future.

I'd be interested to hear your thoughts on this. Have I misunderstood the change? Are there ways to mitigate the risks involved? Do you have a different perspective on this?

mogul:
Chiming in as an outside observer: It doesn't seem like a disruptive change to add this capability (to use Terraform's own resolution/locking) while tolerating the existing syntax, and perhaps emitting a "warning: deprecated syntax, see [url]" warning when it is used. 

I'd like to see this merged sooner rather than later, so I'm more confident that the CSB will be able to take advantage of newer Terraform features and not be harpooned to an older Terraform version that relies on syntax which is unsupported by newer versions of Terraform... There's already a small confidence gap growing between "works as a module with the most recent Terraform" and "can be adapted easily to run in a brokerpak".

mogul:
Hello @blgm (and any other interested folks)... Any thoughts about advancing this work? I'm starting to find that updated versions of modules I'm using in my brokerpak require 0.15+ Terraform versions and HCL syntax.

mogul:
> We wanted to update you that we have released a solution to accept Terraform v0.13 in v0.4.0. We are very keen to know if our solution resolved your issue?

We're trying out 0.4.0 but I stopped short of putting it into production due to [the problem described in Slack](https://cloudfoundry.slack.com/archives/C0164KKEZGX/p1635907093002100).

> Our main concern with this PR is that we want to have reproducible builds that specify the terraform resource versions rather than dynamically getting them.

I think it's fine to keep including all prerequisite binaries and source in the built .brokerpak. However, I want to see the broker move away from requiring an explicit, duplicative list of those binaries in the `manifest.yml` when it can instead rely on [the `.terraform.lock.hcl` file](https://www.terraform.io/docs/language/dependency-lock.html) (supported since Terraform 0.14) for that purpose.

blgm:
Hi @mogul. I think the aim of specifying and including provider versions in a brokerpak is that allows for consistent dependency management. So if you're running a particular brokerpak then you know exactly what code you have on your system, and you can run security scans on the brokerpak. If there's a security vulnerability discovered in a particular version of a dependency then you can move to an updated brokerpak and be sure that the code with the vulnerability is no longer on your system.

I can see how using `.terraform.lock.hcl` could be used as a mechanism to know which provider versions are compatible with a particular service instance. The way I would see it working is that Terraform would download the required providers if they are not already on the system. I can see benefits to that, but I could imagine that some users may see it as problematic as it would make it harder to manage all the dependencies on the system.

I think the two approaches solve slightly different problems for different sets of users. There's definitely merit to both approaches, and I think it's one of those philosophical problems where people will just have different opinions. The approach that we are currently taking to support higher versions of Terraform is designed to follow the existing pattern so that it's a smaller (and less controversial) change.

mogul:
I think I'm not making myself clear... I also want exactly reproducible builds, and I agree that the .brokerpak should include an exact set of binaries known at build time.

What I'm suggesting is that `pak build` should run `terraform init` to download those exact versions specified in the lock file **at build time**. This removes the error-prone chore of manually entering and updating provider version information and download URLs in the manifest.yml file.

blgm:
This was discussed in the CloudFoundry service working group, and we have decided to close it:
- principally this is because it's stale and not moving forwards
- we noted that the original aim of the PR was to support higher versions of Terraform than 0.12. This has now been implemented via another proposal.
- we noted that the developer experience of having to match up Terraform provider versions in the manifest with the Terraform HCL files is not ideal. We may introduce tooling to help with that in the future as part of an epic to improve the developer experience.

As always, we are open to feedback on this. If anyone has a different perspective then please re-open and we can have a discussion.

PR #238, opened 2021-05-11
fix: brokerpak build should fail for duplicate IDs

blgm:
Since https://github.com/cloudfoundry/cloud_controller_ng/commit/8a4926db17807b2cad1736614784f0c57ba49336, Cloud Foundry will fail if a broker uses duplicate IDs for plans as they should be universally unique. This introduces a failure on brokerpak build so that it will no longer be possible to create brokerpaks that will later fail when registering with Cloud Foundry.

[#177727007](https://www.pivotaltracker.com/story/show/177727007)


FelisiaM (approved):
👍

PR #266, opened 2021-07-01
docs: Encryption ADR draft

pivotal-marcela-campo:
[#178658805](https://www.pivotaltracker.com/story/show/178658805)


blgm (approved):
The audience for the document is future developers who are asking the question "why did they do it like this?". This document does answer the question, but it may be longer than many developers have patience to read.

It could also make sense to give the file a meaningful name so that folks don't have to open the file to work out what it's about. `001-database-encryption.md` could be more useful for the situation where there are many files.

FelisiaM (approved):
👍  I really enjoyed reading through the doc and I didn't find it too long. Just enough details to show the options we explored so far and the general strategy.

PR #273, opened 2021-07-21
test: first integration test for DB encryption

blgm:
[#178945167](https://www.pivotaltracker.com/story/show/178945167)


blgm:
This is actually too similar to existing tests: https://github.com/cloudfoundry-incubator/cloud-service-broker/blob/main/brokerapi/brokers/brokers_test.go

PR #275, opened 2021-07-26
Encrypt provision details 178795926

pivotal-marcela-campo:
[#178945167](https://www.pivotaltracker.com/story/show/178945167)


db_service/models/encryption.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"errors"
+	"io"
+)
+
+type GCMEncryptor struct {
+	Key *[32]byte
```
blgm, line 12: Does this need to be public?

db_service/models/encryption_key.go
```diff
@@ -0,0 +1,33 @@
...
+
+var Key [32]byte
+
+// NewKey creates new encryption key
+// TODO: take the key from env vars
+// TODO: Create accessor for Key
+// TODO: Add logging and error handling
+func NewKey() {
+	dbKey := make([]byte, 32)
+	io.ReadFull(rand.Reader, dbKey)
```
blgm, line 31: What happens to the error?

brokerapi/brokers/service_broker.go
```diff
@@ -161,7 +161,10 @@ func (broker *ServiceBroker) Provision(ctx context.Context, instanceID string, d
 	pr := models.ProvisionRequestDetails{
 		ServiceInstanceId: instanceID,
 	}
-	pr.SetRequestDetails(details.RawParameters)
+	err = pr.SetRequestDetails(details.RawParameters)
```
blgm, line 164: Much better to use:
```go
if err := pr.SetRequestDetails(details.RawParameters); err != nil {
```

brokerapi/brokers/service_broker.go
```diff
@@ -596,7 +609,13 @@ func (broker *ServiceBroker) Update(ctx context.Context, instanceID string, deta
 
 	// validate parameters meet the service's schema and merge the user vars with
 	// the plan's
-	vars, err := brokerService.UpdateVariables(instanceID, details, pr.GetRequestDetails(), *plan, request.DecodeOriginatingIdentityHeader(ctx))
+	// TODO
```
blgm, line 612: Who is going to do what, and when?

cmd/serve.go
```diff
@@ -68,11 +71,13 @@ func init() {
 	viper.BindEnv(apiPasswordProp, "SECURITY_USER_PASSWORD")
 	viper.BindEnv(apiPortProp, "PORT")
 	viper.BindEnv(apiHostProp, "CSB_LISTENER_HOST")
+	viper.BindEnv(encryptionKey, "ENCRYPTION_KEY")
```
blgm, line 74: We talked about changing the name of this variable so that it's clear that this is not a completed feature.

db_service/models/db.go
```diff
@@ -29,6 +30,31 @@ const (
...
+
+func ConfigureEncryption(encryptionKey string) Encryptor {
+	encryptor := Encryptor(NewNoopEncryptor())
+	if (strings.TrimSpace(encryptionKey) == encryptionKey) && len(encryptionKey) > 0 {
+		key := []byte(encryptionKey)
+		var keyAs32ByteArray [32]byte
+		copy(keyAs32ByteArray[:], key)
+		encryptor = NewGCMEncryptor(&keyAs32ByteArray)
+	}
+	// TODO else should probably return an error to be on the safe side
```
blgm, line 47: Who will do this, and when?

db_service/models/encryption_key.go
pivotal-marcela-campo: I am removing this method, we just use it in tests at the moment.

db_service/models/encryption.go
pivotal-marcela-campo: At the moment we are using it from one of the test to check we are creating the right encryptor and setting up the right key.

db_service/models/db.go
pivotal-marcela-campo: Part of the validation story. Will remove the comment.

PR #278, opened 2021-07-29
encryption: encode before saving to database to avoid mysql encoding

pivotal-marcela-campo:
restrictions

MySQL's utf8 permits only the Unicode characters that can be represented
with 3 bytes in UTF-8, Changing the encoding in the dabatase fixes it,
but it requires migrations and it is also database dependent.
Encoding to b64 makes it more straightforward.

[#178795926](https://www.pivotaltracker.com/story/show/178795926)

### DIFF
--- a/db_service/models/db_test.go
+++ b/db_service/models/db_test.go
@@ -47,13 +47,15 @@ var _ = Describe("Db", func() {
 
 					credentials := models.ServiceBindingCredentials{}
 					err := credentials.SetOtherDetails(otherDetails)
-					Expect(err).NotTo(HaveOccurred())
+					Expect(err).ToNot(HaveOccurred())
 
 					By("checking that it's no longer in plaintext")
-					Expect(credentials.OtherDetails).NotTo(Equal(expectedJSON))
+					Expect(credentials.OtherDetails).ToNot(Equal(expectedJSON))
 
 					By("being able to decrypt to get the value")
-					Expect(encryptor.Decrypt([]byte(credentials.OtherDetails))).To(Equal([]byte(expectedJSON)))
+					decrypted, err := encryptor.Decrypt(credentials.OtherDetails)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(string(decrypted)).To(Equal(expectedJSON))
 				})
 			})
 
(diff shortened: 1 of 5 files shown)


db_service/models/db_test.go
```diff
@@ -358,14 +362,14 @@ var _ = Describe("Db", func() {
 						F func()
 					}{F: func() {}})
 
-					Expect(err).ToNot(BeNil(), "Should have returned an error")
+					Expect(err).To(HaveOccurred(), "Should have returned an error")
```
blgm, line 365: Two questions:
1) Is the default negative message for `HaveOccurred()` not good enough?
2) Would it not be better to assert for a specific error?

db_service/models/db_test.go
```diff
@@ -47,13 +47,15 @@ var _ = Describe("Db", func() {
 
 					credentials := models.ServiceBindingCredentials{}
 					err := credentials.SetOtherDetails(otherDetails)
-					Expect(err).NotTo(HaveOccurred())
+					Expect(err).ToNot(HaveOccurred())
```
blgm, line 50: There seems to be a concerted effort to replace all the "NotTo"s with "ToNot". What are the reasons for this?

internal/encryption/encryption.go
```diff
@@ -16,31 +17,37 @@ func NewGCMEncryptor(key *[32]byte) GCMEncryptor {
...
+	// Return the encrypted text appended to the nonce, encoded in b64
+	sealed := gcm.Seal(nonce, nonce, plaintext, nil)
+	return b64.StdEncoding.EncodeToString(sealed), nil
 }
 
-func (d GCMEncryptor) Decrypt(ciphertext []byte) (plaintext []byte, err error) {
+func (d GCMEncryptor) Decrypt(ciphertext string) ([]byte, error) {
+	decoded, err := b64.StdEncoding.DecodeString(ciphertext)
+	if err != nil {
+		return nil, err
```
blgm, line 48: If this error handling is removed then no tests fail. I'm not saying that we must test this error handling - I just wanted to make sure it was a choice rather than an accident.

blgm (approved):

internal/encryption/encryption.go
pivotal-marcela-campo: Yes, we are mocking the encryptor for the db_tests tests, so we have the choice of either trying to cause this in an integration test or unit test the encryptor. There is a story top of backlog to unit test the encryptor so I just deferred to that point for convenience.

db_service/models/db_test.go
pivotal-marcela-campo: We seemed to be consistently using ToNot in other test files. We were also using ToNot for Equals. And I started using ToNot for new error assertions and found later that this particular file had NotTo. So just decided on one and changed the others.

PR #283, opened 2021-08-16
Encryption tests

pivotal-marcela-campo:
Testing encryption on update, unbind and deprovision for both terraform and params data

[#179006636](https://www.pivotaltracker.com/story/show/179006636)

### DIFF
--- a/integrationtest/database_encryption_test.go
+++ b/integrationtest/database_encryption_test.go
@@ -23,15 +23,18 @@ import (
 
 var _ = Describe("Database Encryption", func() {
 	const (
-		provisionParams      = `{"foo":"bar"}`
-		bindParams           = `{"baz":"quz"}`
-		provisionOutput      = `{"provision_output":"provision output value"}`
-		provisionOutputValue = `value = \"provision output value\"`
-		bindOutput           = `{"bind_output":"provision output value and bind output value"}`
-		bindOutputValue      = `value = \"${var.provision_output} and bind output value\"`
-		tfStateKey           = `"tfstate":`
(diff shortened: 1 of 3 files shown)


integrationtest/database_encryption_test.go
```diff
@@ -49,54 +55,60 @@ var _ = Describe("Database Encryption", func() {
 		serviceBindingGUID  string
 	)
 
-	persistedRequestDetails := func() string {
+	findRecord := func(dest interface{}, query, guid string) {
```
blgm, line 58: Nice improvement

integrationtest/database_encryption_test.go
```diff
@@ -204,17 +267,45 @@ var _ = Describe("Database Encryption", func() {
...
+
+			By("ckecking the service instance fields after deprovision", func() {
+				deprovisionServiceInstance()
+				expectServiceInstanceDetailsToNotExist()
+				Expect(persistedServiceInstanceTerraformWorkspace()).NotTo(SatisfyAny(
+					ContainSubstring(provisionOutputStateValue),
+					ContainSubstring(updateOutputStateValue),
+					ContainSubstring(tfStateKey),
+				))
+			})
```
blgm, line 308: The two tests are opposites of each other. I've checked them and the look to be correct. They are getting long enough that it's hard to check that they really are opposites of each other. If we change them again, we should try to find another way to express them so that it's easier to check that they are opposites.

blgm (approved):

integrationtest/database_encryption_test.go
pivotal-marcela-campo: good call. Will keep that in mind

PR #287, opened 2021-08-24
Detect duplicate plan ids #179042292

FelisiaM:
[#179042292](https://www.pivotaltracker.com/story/show/179042292)


pkg/broker/broker_test.go
```diff
@@ -73,25 +73,15 @@ func ExampleServiceDefinition_GetPlanById() {
...
-	fmt.Printf("builtin-plan: %q %v\n", plan.Name, err)
-
-	plan, err = service.GetPlanById("custom-plan")
-	fmt.Printf("custom-plan: %q %v\n", plan.Name, err)
+	plan, err := service.GetPlanById("test-plan")
+	fmt.Printf("test-plan: %q %v\n", plan.Name, err)
 
 	_, err = service.GetPlanById("missing-plan")
 	fmt.Printf("missing-plan: %s\n", err)
-
```
blgm, line 91: This is an example rather than a test, so I'd expect to see the expected output.
https://go.dev/blog/examples

integrationtest/catalog_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+		err := os.Chdir(originalDir)
+		Expect(err).NotTo(HaveOccurred())
+
+		err = os.RemoveAll(workDir)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	When("a service offering has duplicate plan IDs", func() {
+		JustBeforeEach(func() {
```
blgm, line 77: Nested `JustBeforeEach()` is likely to get confusing... Can we just have one?

blgm (changes requested):

integrationtest/database_encryption_test.go
```diff
@@ -170,6 +170,7 @@ var _ = Describe("Database Encryption", func() {
 		brokerPort = freePort()
 		databaseFile = path.Join(workDir, "databaseFile.dat")
 		runBrokerCommand := exec.Command(csb, "serve")
+		os.Unsetenv("CH_CRED_HUB_URL")
```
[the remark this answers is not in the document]
FelisiaM: If you have an env targeted the credhub env vars are set. That causes the test to fail (as it tries to call CH) - I think there might be better ways to do this but the idea is to clean up before setting the test but that's the first I thought about.

integrationtest/catalog_test.go
FelisiaM: 👍

integrationtest/catalog_test.go
```diff
@@ -0,0 +1,127 @@
...
+		session, err := Start(buildBrokerpakCommand, GinkgoWriter, GinkgoWriter)
+		Expect(err).NotTo(HaveOccurred())
+		Eventually(session, time.Minute).Should(Exit(0))
+
+		brokerUsername = uuid.New()
+		brokerPassword = uuid.New()
+		brokerPort = freePort()
+		databaseFile = path.Join(workDir, "databaseFile.dat")
+		runBrokerCommand = exec.Command(csb, "serve")
+		os.Unsetenv("GSB_SERVICE_ALPHA_SERVICE_PLANS")
```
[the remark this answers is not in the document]
FelisiaM: It was left over from before I changed the test. Removed it now

integrationtest/catalog_test.go
```diff
@@ -0,0 +1,127 @@
...
+		os.Unsetenv("GSB_SERVICE_BETA_SERVICE_PLANS")
+		runBrokerCommand.Env = append(
```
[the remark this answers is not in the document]
FelisiaM: Nope - I removed it

pkg/broker/broker_test.go
FelisiaM: I added it back!

PR #288, opened 2021-08-25
Multiple encryption pass #179324852

FelisiaM:

PR #289, opened 2021-08-31
Validations for new password style

blgm:
In future we plan to accept multiple encryption passwords with format:
```json
[
  {
    "label": "my-label",
    "primary": true,
    "password": {
      "secret": "fake-super-secret-password"
    }
  },
  {...}
]
```
The format is based on the output of the tile configuration

Based on PR #288 by @FelisiaM

### DIFF
--- a/internal/encryption/parser/parser.go
+++ b/internal/encryption/parser/parser.go
@@ -1,0 +1,1 @@
+package configparser
--- a/internal/encryption/parser/parser.go
+++ b/internal/encryption/parser/parser.go
@@ -1,1 +1,88 @@
-package configparser
+package parser
+
+import (
+	"encoding/json"
+	"fmt"
+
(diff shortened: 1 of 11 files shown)


internal/encryption/parser/parser.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/cloudfoundry-incubator/cloud-service-broker/pkg/validation"
+)
+
+type PasswordEntry struct {
+	Label   string `jsonry:"label"`
```
kirederik, line 11: Any benefit in using jsonry in this case?

kirederik (approved):

internal/encryption/parser/parser.go
blgm: No! It's from something I tried that didn't work. Removed it now.

PR #290, opened 2021-09-01
feat: encryption passwords (experimental)

blgm:
Previously to enable encryption, the 32 byte key had to be specified.
Now a password is used instead, and this may be of any (appropriate)
length.
Future work will allow for migration between passwords.

This is inspired by #288 by @FelisiaM 

[#179324852](https://www.pivotaltracker.com/story/show/179324852)


db_service/models/historical_db.go
```diff
@@ -261,3 +261,18 @@ type TerraformDeploymentV2 struct {
 func (TerraformDeploymentV2) TableName() string {
 	return "terraform_deployments"
 }
+
+// PasswordMetadataV1 contains information about the passwords, but never the
+// passwords themselves
+type PasswordMetadataV1 struct {
+	gorm.Model
+
+	Label   string `gorm:"index;unique;not null"`
```
kirederik, line 270: The TAS tile generates a guid for the cloud controller encryption keys, and my guess is that it uses that guid to identify the key. Users can then change labels at will. 

It looks like we opted for identifying the password by a label, which in turn makes the label immutable. That's fine, just wanted to highlight the difference.

kirederik (approved):
LGTM

PR #293, opened 2021-09-04
chore: fix some warnings from updated Gorm

blgm:
Looks like the newer Gorm is noisier about bad usage such as using a
SELECT to get a record count. These changes fix the warnings that were
printed out during some integration tests.

[#177268286](https://www.pivotaltracker.com/story/show/177268286)

### DIFF
--- a/pkg/providers/tf/job_runner.go
+++ b/pkg/providers/tf/job_runner.go
@@ -69,40 +69,33 @@ type TfJobRunner struct {
 func (runner *TfJobRunner) StageJob(ctx context.Context, jobId string, workspace *wrapper.TerraformWorkspace) error {
 	workspace.Executor = runner.Executor
 
-	workspaceString, err := workspace.Serialize()
-	if err != nil {
-		return err
-	}
-
 	exists, err := db_service.ExistsTerraformDeploymentById(ctx, jobId)
 	if err != nil {
 		return err
 	}
 
+	deployment := &models.TerraformDeployment{ID: jobId}
 	if exists {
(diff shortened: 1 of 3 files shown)


pkg/providers/tf/job_runner.go
```diff
@@ -74,25 +74,35 @@ func (runner *TfJobRunner) StageJob(ctx context.Context, jobId string, workspace
...
+		}
+
 		if err := deployment.SetWorkspace(workspaceString); err != nil {
 			return err
 		}
 
 		deployment.LastOperationType = "validation"
 		return runner.operationFinished(nil, workspace, deployment)
-	}
+	} else {
```
pivotal-marcela-campo, line 94: Do we need the `else` here? we are returning inside the `if` block

pkg/providers/tf/job_runner.go
```diff
@@ -74,25 +74,35 @@ func (runner *TfJobRunner) StageJob(ctx context.Context, jobId string, workspace
...
+		}
 
-	deployment := &models.TerraformDeployment{
-		ID:                jobId,
-		LastOperationType: "validation",
-	}
-	if err := deployment.SetWorkspace(workspaceString); err != nil {
-		return err
+		db_service.CreateTerraformDeployment(context.Background(), deployment)
+		return runner.operationFinished(nil, workspace, deployment)
```
pivotal-marcela-campo, line 104: both if and else return the same thing. So maybe we can remove the in block returns and have the generic one at the end instead

pivotal-marcela-campo (approved):

PR #294, opened 2021-09-06
feat: migration from plaintext to encrypted data

blgm:
If a database has been using plaintext, when a new primary password is specified, the database will be re-encrypted on the next startup.

[#179355268](https://www.pivotaltracker.com/story/show/179355268)


internal/encryption/dbrotator/dbrorator.go
```diff
@@ -0,0 +1,103 @@
...
+		encryptServiceBindingCredentials,
+		encryptTerraformWorkspaces,
+	}
+	for _, e := range dbEncryptors {
+		if err := e(db); err != nil {
+			return nil
+		}
+	}
+
+	return nil
```
FelisiaM, line 22: At the end of re-encryption we still have not updated the new password to be "primary" in the DB. This means that when the app restarts it will attempt to re-encrypt again even though nothing will have changed.

integrationtest/database_encryption_test.go
```diff
@@ -205,108 +223,159 @@ var _ = Describe("Database Encryption", func() {
...
+			serviceBindingGUID := uuid.New()
+			createBinding(serviceInstanceGUID, serviceBindingGUID)
+			createBinding(serviceInstanceGUID, serviceBindingGUID)
+			Expect(persistedServiceBindingDetails(serviceInstanceGUID)).To(bePlaintextBindingDetails)
+			Expect(persistedServiceBindingTerraformWorkspace(serviceInstanceGUID, serviceBindingGUID)).To(bePlaintextBindingTerraformState)
+
+			By("restarting the broker with a password")
+			brokerSession.Terminate()
+			const encryptionPasswords = `[{"primary":true,"label":"my-password","password":{"secret":"supersecretcoolpassword"}}]`
+			brokerSession = startBroker(true, encryptionPasswords)
```
FelisiaM, line 363: I think there is one more case that this test should check. If you restart the broker again with the same password configuration it should be able to access the same data and delete instances/bindings. This would prove that we are keeping track of something relating to the password in the DB.

FelisiaM (changes requested):

internal/encryption/dbrotator/dbrorator_test.go
```diff
@@ -0,0 +1,108 @@
+package dbrotator_test
```
FelisiaM, line 1: same issue with typo in the file name

internal/encryption/passwordcombiner/combinedpasswords.go
```diff
@@ -0,0 +1,34 @@
+package passwordcombiner
+
+import "github.com/cloudfoundry-incubator/cloud-service-broker/internal/encryption/gcmencryptor"
+
+type CombinedPassword struct {
+	Label         string
+	Secret        string
+	Salt          []byte
+	Encryptor     gcmencryptor.GCMEncryptor
+	parsedPrimary bool
```
FelisiaM, line 10: I wonder if there is a better name for `parsedPrimary` since "parsing" is not really the important concept here. It's more "configured".  What do you think?

FelisiaM (approved):

internal/encryption/dbrotator/dbrorator.go
blgm: Added.

integrationtest/database_encryption_test.go
blgm: Added.

PR #295, opened 2021-09-08
chore: encryption ADR doc to reflect recent decisions

blgm:
[#179324852](https://www.pivotaltracker.com/story/show/179324852)


docs/arch/adr-001.md
```diff
@@ -26,97 +26,102 @@ We are addressing the need to encrypt this data.
...
+
+## Code structure
+
+The GORM package that is used access the database allows for custom types that implement
+the Scanner and Valuer interfaces which can transparently transform data as it is saved or
+loaded from the database. This can be used to JSON encode/decode data, or to encrypt/decrypt it.
+We decided not to use this approach as there was already a precedent for using accessor methods
+on objects to do JSON encoding/decoding, so we enhanced those accessor methods to also encrypt
+and decrypt the data. At some future point we may refactor the accessor methods to implement
+Scanner/Valuer and make the data transformation transparent.
```
pivotal-marcela-campo, line 114: I am just wondering if we should keep this fragment somehow, to have all the reasoning behind why we chose one approach over the other in writing. 

```
We will use `3. Accessor methods` to perform encryption/decryption. The reasons behind it are:
* It makes clear when and where credentials are being handled in plain text, as opposed of a custom datatype that would only be made visible in the model definition as an annotation.
* Decryption can be left out until credentials are actually needed and will be perform as soon as they are set, limiting the possibilities of leaking them with logging, under bad error handling etc. If we were to use the custom data type approach, encryption would happen as soon as we read from the database and delayed until we actually write to the database.

We will rename variables in the code to make clear which fields can contain sensitive data (as opposed to `other_details`)

We will consider renaming database fields.    
```

PR #296, opened 2021-09-09
feat: Decrypt DB when turning encryption off

FelisiaM:
[#179355269](https://www.pivotaltracker.com/story/show/179355269)


integrationtest/database_encryption_test.go
```diff
@@ -391,4 +391,60 @@ var _ = Describe("Database Encryption", func() {
...
+			provisionServiceInstance(serviceInstanceGUID)
+			Expect(persistedRequestDetails(serviceInstanceGUID)).NotTo(bePlaintextProvisionParams)
+			Expect(persistedServiceInstanceDetails(serviceInstanceGUID)).NotTo(bePlaintextProvisionOutput)
+			Expect(persistedServiceInstanceTerraformWorkspace(serviceInstanceGUID)).NotTo(haveAnyPlaintextServiceTerraformState)
+
+			By("creating a binding and checking the fields are stored encrypted")
+			serviceBindingGUID := uuid.New()
+			createBinding(serviceInstanceGUID, serviceBindingGUID)
+			Expect(persistedServiceBindingDetails(serviceInstanceGUID)).NotTo(bePlaintextBindingDetails)
+			Expect(persistedServiceBindingTerraformWorkspace(serviceInstanceGUID, serviceBindingGUID)).NotTo(bePlaintextBindingTerraformState)
```
blgm, line 416: Should use `haveAnyPlaintextBindingTerraformState` matcher

integrationtest/database_encryption_test.go
```diff
@@ -391,4 +391,60 @@ var _ = Describe("Database Encryption", func() {
...
+				Say(`cloud-service-broker.rotating-database-encryption\S*"data":{"new-primary":"my-password","previous-primary":"none"}}`),
+				Say(`cloud-service-broker.database-encryption\S*"data":{"primary":"my-password"}}`),
+			))
+
+			By("creating a service instance and checking fields are stored encrypted")
+			serviceInstanceGUID := uuid.New()
+			provisionServiceInstance(serviceInstanceGUID)
+			Expect(persistedRequestDetails(serviceInstanceGUID)).NotTo(bePlaintextProvisionParams)
+			Expect(persistedServiceInstanceDetails(serviceInstanceGUID)).NotTo(bePlaintextProvisionOutput)
+			Expect(persistedServiceInstanceTerraformWorkspace(serviceInstanceGUID)).NotTo(haveAnyPlaintextServiceTerraformState)
```
blgm, line 410: Should use `haveAnyPlaintextServiceTerraformState` matcher

integrationtest/database_encryption_test.go
```diff
@@ -391,4 +391,60 @@ var _ = Describe("Database Encryption", func() {
...
+			Expect(brokerSession.Out).To(SatisfyAll(
+				Say(`cloud-service-broker.rotating-database-encryption\S*"data":{"new-primary":"none","previous-primary":"my-password"}}`),
+				Say(`cloud-service-broker.database-encryption\S*"data":{"primary":"none"}}`),
+			))
+
+			By("checking that the previous data is now decrypted")
+			Expect(persistedRequestDetails(serviceInstanceGUID)).To(bePlaintextProvisionParams)
+			Expect(persistedServiceInstanceDetails(serviceInstanceGUID)).To(bePlaintextProvisionOutput)
+			Expect(persistedServiceInstanceTerraformWorkspace(serviceInstanceGUID)).To(haveAnyPlaintextServiceTerraformState)
+			Expect(persistedServiceBindingDetails(serviceInstanceGUID)).To(bePlaintextBindingDetails)
```
blgm, line 431: Should use `bePlaintextBindingTerraformState` matcher

integrationtest/database_encryption_test.go
```diff
@@ -391,4 +391,60 @@ var _ = Describe("Database Encryption", func() {
...
+			Expect(persistedServiceBindingTerraformWorkspace(serviceInstanceGUID, serviceBindingGUID)).To(haveAnyPlaintextBindingTerraformState)
+
+			By("restarting the broker with encryption turned off again")
+			brokerSession.Terminate()
+			encryptionPasswords = `[{"primary":false,"label":"my-password","password":{"secret":"supersecretcoolpassword"}}]`
```
blgm, line 436: Would it be a better test to remove the password completely?

internal/encryption/parse_configuration.go
```diff
@@ -28,28 +28,35 @@ func ParseConfiguration(db *gorm.DB, enabled bool, passwords string) (Configurat
...
+			encryptor = noopencryptor.New()
+		} else {
+			encryptor = parsedPrimary.Encryptor
 		}
 
-		rotationEncyptor = compoundencryptor.New(parsedPrimary.Encryptor, secondaryEncryptors...)
+		rotationEncyptor = compoundencryptor.New(encryptor, decryptors...)
+	}
+
+	switch {
```
blgm, line 53: I think this switch should be split, so that the guard clauses are at the top of the function:
```go
switch {
case enabled && !parsedPrimaryOK:
  return Configuration{}, errors.New("encryption is enabled but no primary password is set")
case !enabled && parsedPrimaryOK:
  return Configuration{}, errors.New("encryption is disabled but a primary password is set")
}
```

And the success branches should have duplication removed:
```go
result := Configuration{
  ConfiguredPrimaryLabel: labelName(parsedPrimary.Label),
  StoredPrimaryLabel: labelName(storedPrimary.Label),
  Changed: changed,
  RotationEncryptor: rotationEncyptor,
}

if !enabled && !parsedPrimaryOK {
  result.Encryptor = noopencryptor.New()
} else {
  result.Encryptor = parsedPrimary.Encryptor
}

return result, nil
```

internal/encryption/update_password_metadata.go
```diff
@@ -25,7 +25,7 @@ func UpdatePasswordMetadata(db *gorm.DB, configuredPrimaryLabel string) error {
 			}
 		}
 
-		if configuredPrimaryLabel != "" {
+		if !(configuredPrimaryLabel == "" || configuredPrimaryLabel == "none") {
```
blgm, line 28: The `"none"` text should really only be a UI external. We should display it to a user, but not use it for internal processing. Maybe we should move the conversion to the `serve()` function.

internal/encryption/compoundencryptor/compoundencryptor.go
```diff
@@ -2,23 +2,23 @@ package compoundencryptor
...
+	decryptors []Encryptor
 }
 
 func (c CompoundEncryptor) Encrypt(plaintext []byte) (string, error) {
-	return c.primary.Encrypt(plaintext)
+	return c.encryptor.Encrypt(plaintext)
 }
 
 func (c CompoundEncryptor) Decrypt(ciphertext string) (data []byte, err error) {
-	for _, decryptor := range append([]Encryptor{c.primary}, c.secondaries...) {
```
blgm, line 20: I think change introduces a bug. In cases where the database re-encryption fails without completing, when the CSB restarts it will try to re-encrypt the database again. It therefore needs to be able to decrypt with the primary otherwise it will not be able to recover.

blgm (changes requested):

internal/encryption/update_password_metadata.go
FelisiaM: Good catch - I put it as a temp solution and meant to change that before submitting the PR but forgot.

internal/encryption/compoundencryptor/compoundencryptor.go
FelisiaM: I added a [test](https://github.com/cloudfoundry-incubator/cloud-service-broker/pull/296/files#diff-67301087574d31ce4500119d0c090bdf29419a445babc3be72495823f7241ec2R180) for this use case and I think there is no bug.  In cases where the database re-encryption fails without completing, the intended new key would still be stored in the DB, just not labeled as "primary". This means it will be part of the combined passwords that are returned from [`passwordcombiner.CombineWithStoredMetadata`](https://github.com/cloudfoundry-incubator/cloud-service-broker/blob/4ac02be2930f18349350da986d2ad4f2bb997385/internal/encryption/parse_configuration.go#L23). Since we add the encryptors from the combined passwords as decryptors for rotation the right decryptor will be available.

blgm: Ok. A test is definitely the right way to resolve this.

blgm (approved):

PR #300, opened 2021-09-14
feat: Clean up password metadata that are no longer used

FelisiaM:
[#179355270](https://www.pivotaltracker.com/story/show/179355270)

### DIFF
--- a/integrationtest/database_encryption_test.go
+++ b/integrationtest/database_encryption_test.go
@@ -38,6 +38,7 @@ var _ = Describe("Database Encryption", func() {
 		serviceInstanceFKQuery    = "service_instance_id = ?"
 		serviceInstanceIdQuery    = "id = ?"
 		tfWorkspaceIdQuery        = "id = ?"
+		passwordMetadataQuery     = "label = ?"
 	)
 
 	var (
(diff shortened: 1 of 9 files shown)


integrationtest/database_encryption_test.go
```diff
@@ -384,6 +410,9 @@ var _ = Describe("Database Encryption", func() {
 			Expect(string(brokerSession.Out.Contents())).NotTo(ContainSubstring(`cloud-service-broker.rotating-database-encryption`))
 			Expect(brokerSession.Out).To(Say(`cloud-service-broker.database-encryption\S*"data":{"primary":"my-password"}}`))
 
+			By("checking that the password metadata are still stored")
```
kirederik, line 413: When would this test fail? What has changed since line 397?

integrationtest/database_encryption_test.go
```diff
@@ -439,6 +471,9 @@ var _ = Describe("Database Encryption", func() {
 			Expect(string(brokerSession.Out.Contents())).NotTo(ContainSubstring(`cloud-service-broker.rotating-database-encryption`))
 			Expect(brokerSession.Out).To(Say(`cloud-service-broker.database-encryption\S*"data":{"primary":"none"}}`))
 
+			By("checking that the password metadata are removed")
```
kirederik, line 474: I don't get what's happening before this line is reached that's causing the password_metadata table to be wiped of all passwords 🤔

integrationtest/database_encryption_test.go
```diff
@@ -484,6 +522,10 @@ var _ = Describe("Database Encryption", func() {
 				Say(`cloud-service-broker.database-encryption\S*"data":{"primary":"my-second-password"}}`),
 			))
 
+			By("checking that the password metadata are stored")
```
kirederik, line 525: I don't get what's happening before this line is reached that's causing the password_metadata table to be updated 🤔

internal/encryption/delete_password_metadata_test.go
```diff
@@ -0,0 +1,88 @@
...
+				}).Error).NotTo(HaveOccurred())
+				Expect(db.Create(&models.PasswordMetadata{
+					Label:   "other-to-delete",
+					Salt:    []byte("random-salt"),
+					Canary:  "test-value",
+					Primary: false,
+				}).Error).NotTo(HaveOccurred())
+			})
+
+			It("all matching password metadata should be deleted", func() {
```
kirederik, line 52: It all matching password metadata should be deleted doesn't read nicely

internal/encryption/parse_configuration.go
```diff
@@ -71,3 +85,28 @@ func ParseConfiguration(db *gorm.DB, enabled bool, passwords string) (Configurat
 
 	return result, nil
 }
+
+func toDelete(stored []models.PasswordMetadata, configured []passwordparser.PasswordEntry) []string {
```
kirederik, line 89: A different and more time efficient algorithm could use a hash map (prob irrelevant for the problem space tbh; untested code below):

```go
var configuredMap map[string]bool

for _, c := range configured {
        configuredMap[c.Label] = true
}

for _, s := range stored {
        if !configuredMap[s.Label] {
                 toDelete = append(toDelete, s.Label)
        }
}

return toDelete
```

internal/encryption/passwordcombiner/combiner.go
```diff
@@ -12,13 +12,8 @@ import (
...
 )
 
-func CombineWithStoredMetadata(db *gorm.DB, passwords string) (CombinedPasswords, error) {
-	parsed, err := passwordparser.Parse(passwords)
-	if err != nil {
-		return nil, err
-	}
-
-	stored, storedPrimary, err := loadPasswordMetadata(db)
+func Combine(db *gorm.DB, parsed []passwordparser.PasswordEntry, storedPassMetadata []models.PasswordMetadata) (CombinedPasswords, error) {
```
kirederik, line 15: This function confuses me. It does so many things and I'm not sure passwordcombiner.Combine conveys that.

Note: not a problem introduced by this PR

kirederik (commented):
Some comments below. Most of it were not necessarily introduced by this PR though

PR #324, opened 2021-11-12
Dynamic hcl 179924065

pivotal-marcela-campo:
This PR updates the HCL stored in the database for existing SI/bindings prior to executing update/unbind/deprovision.

Ideally we would not store it at all, and this is a  first step towards that while keeping it backwards compatible.


pkg/providers/tf/provider.go
```diff
@@ -88,12 +146,22 @@ func (provider *terraformProvider) Update(ctx context.Context, provisionContext
...
 	}
 
-	err := provider.jobRunner.Update(ctx, tfId, provisionContext.ToMap())
+	err := workspaceUpdator.UpdateWorkspaceHCL(ctx, provider.serviceDefinition.ProvisionSettings, provisionContext, tfId)
+	if err != nil {
+		return models.ServiceInstanceDetails{}, err
+	}
 
-	return models.ServiceInstanceDetails{
+	// if err = workspace.Validate(); err != nil {
```
blgm, line 154: Ideally this would be cleaned up

pkg/providers/tf/provider.go
```diff
@@ -200,6 +268,11 @@ func (provider *terraformProvider) Unbind(ctx context.Context, instanceRecord mo
 		"tfId":     tfId,
 	})
 
+	err := workspaceUpdator.UpdateWorkspaceHCL(ctx, provider.serviceDefinition.BindSettings, vc, tfId)
```
blgm, line 271: I have a slight preference for:
```go
if err := work....; err != nil {
```
Because it reduces the scope of `err` and makes it harder to use the wrong `err` later in the block, but this is a very minor point.

blgm (approved):

PR #350, opened 2022-01-25
chore: allow disabling of parameters check

blgm:
Because we have added a new check to provision and update parameters
it's possible that some workflows that used to pass will now fail. While
the correct action is probably to update the respective brokerpaks,
allowing the check to be turned off give users options to delay those
updates until an appropriate moment.

[#178067889](https://www.pivotaltracker.com/story/show/178067889)


FelisiaM (approved):
👍

PR #362, opened 2022-01-28
feat: Add error message when updating plan config

FelisiaM:
If a property has been defined on the plan properties (when creating a plan) that property cannot be updated by the user in provision or update time.

[#181013270](https://www.pivotaltracker.com/story/show/181013270)


brokerapi/brokers/service_broker.go
```diff
@@ -680,10 +683,36 @@ func validateParameters(rawParams json.RawMessage, validUserInputFields []broker
...
+	if len(rawParams) == 0 {
 		return nil
 	}
 
+	var params map[string]interface{}
+	if err := json.Unmarshal(rawParams, &params); err != nil {
+		return ErrInvalidUserInput
+	}
+
+	err := validatePlanDefinedOverrides(params, plan)
```
pivotal-marcela-campo, line 702: I find the name of this function a bit confusing. It is not validating overrides but checking there are no attempted overrides. I cannot though think of a name that starts with `validate`. `validateNoPlanDefinedProperties`? not sure it is better. `checkNoPlanDefinedProperties`? we are already inside a validate function here so maybe we can drop the prefix for the helper function.

brokerapi/brokers/service_broker.go
```diff
@@ -706,6 +735,21 @@ func validateParameters(rawParams json.RawMessage, validUserInputFields []broker
 	return fmt.Errorf("additional properties are not allowed: %s", strings.Join(invalidParams, ", "))
 }
 
+func validatePlanDefinedOverrides(params map[string]interface{}, plan *broker.ServicePlan) error {
+	var invalidPlanParams []string
```
pivotal-marcela-campo, line 739: I would call this just `serviceParams`, it is not so clear at the beginning of the function why they would be invalid. Until we read the error message. I actually though the condition in line `741` was inverted.

brokerapi/brokers/brokers_test.go
```diff
@@ -421,6 +425,15 @@ func TestServiceBroker_Provision(t *testing.T) {
...
+		"bad-request-invalid-plan-defined-parameter": {
+			ServiceState: StateNone,
+			Check: func(t *testing.T, broker *ServiceBroker, stub *serviceStub, encryptor *storagefakes.FakeEncryptor) {
+				req := stub.ProvisionDetails()
+				req.RawParameters = json.RawMessage(`{"foo":"bar","plan-defined-key":42,"other-plan-defined-key":"test","other_invalid":false}`)
+				_, err := broker.Provision(context.Background(), fakeInstanceId, req, true)
+				assertEqual(t, "errors should match", errors.New("plan defined properties cannot be changed: other-plan-defined-key, plan-defined-key"), err)
+			},
+		},
 		"bad-request-invalid-parameter-disabled": {
```
pivotal-marcela-campo, line 437: It looks like this scenario is only tested here, should we test for `bind` as well?

pivotal-marcela-campo (changes requested):

brokerapi/brokers/brokers_test.go
FelisiaM: This code change does not impact bind operation. The plan properties are not merged similarly for [bind](https://github.com/cloudfoundry-incubator/cloud-service-broker/blob/9c3efd6cfa05133b783dbd6bbdaf1a7deafa9648/pkg/broker/service_definition.go#L462) as for [provision or update](https://github.com/cloudfoundry-incubator/cloud-service-broker/blob/9c3efd6cfa05133b783dbd6bbdaf1a7deafa9648/pkg/broker/service_definition.go#L392)

brokerapi/brokers/service_broker.go
FelisiaM: I dont think I agree. The only usage of the variable is to hold the list of the invalid params regardless of when/how it gets populated.

brokerapi/brokers/service_broker.go
FelisiaM: We might be getting into semantics here but any validation can be expressed as a check. So maybe a way to find a better name is:
 `validateProvisionParameters` = `checkProvisionParametersAreValid` so 
`validatePlanDefinedOverrides` = `checkPlanDefinedParametersAreNotOverriden` = `validateNoPlanParametersOverrides`

I also think that each "validate" func can be used independently and they can contain the "validate" word.

pivotal-marcela-campo: I don't have a problem with the `validate` world specifically, more of a finding an appropriate name. The renamed version makes much more sense to me. Thanks.

pivotal-marcela-campo (approved):

PR #371, opened 2022-02-03
test: use Ginkgo TempDir() rather than os.MkdirTemp()

blgm:
The advantage of the Ginkgo helper is that errors are automatically
checked and the directory is automatically cleaned up.

### DIFF
--- a/integrationtest/catalog_test.go
+++ b/integrationtest/catalog_test.go
@@ -16,7 +16,6 @@ import (
 )
 
 var _ = Describe("Catalog", func() {
-
 	var (
 		originalDir      string
 		fixturesDir      string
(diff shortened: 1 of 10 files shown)


integrationtest/catalog_test.go
```diff
@@ -35,9 +34,8 @@ var _ = Describe("Catalog", func() {
 		Expect(err).NotTo(HaveOccurred())
 		fixturesDir = path.Join(originalDir, "fixtures", "brokerpak-for-catalog-test")
 
-		workDir, err = os.MkdirTemp("", "*-csb-test")
-		Expect(err).NotTo(HaveOccurred())
-		err = os.Chdir(workDir)
+		workDir = GinkgoT().TempDir()
+		os.Chdir(workDir)
```
jimbo459, line 38: We're not checking the return error value of `os.Chdir` here, can we add that assignment?

PR #386, opened 2022-02-15
chore: Replace with provision test with ginkgo test

FelisiaM:
### Checklist:

* [ ] Have you added Draft Release Notes in `docs/draft-release-notes.md`? not applicable


brokerapi/broker/provision_test.go
```diff
@@ -0,0 +1,310 @@
...
+
+			By("validating response")
+			Expect(response.IsAsync).To(BeTrue())
+			Expect(response.DashboardURL).To(BeEmpty())
+			Expect(response.OperationData).To(Equal(operationID))
+
+			By("validating provider provision has been called")
+			Expect(fakeServiceProvider.ProvisionCallCount()).To(Equal(1))
+			actualContext, actualVars := fakeServiceProvider.ProvisionArgsForCall(0)
+			Expect(actualContext.Value(middlewares.OriginatingIdentityKey)).To(Equal(expectedHeader))
```
pivotal-marcela-campo, line 131: This is not testing the same as the original test.  The original test "originating-header" would test that the decoded object with values was merged into the provision variables. This is just testing that we are passing the same context object to provision that we received in the api call.

brokerapi/broker/provision_test.go
```diff
@@ -0,0 +1,310 @@
...
+				fakeStorage.StoreServiceInstanceDetailsReturns(errors.New("failed to store SI details"))
+			})
+
+			It("should error", func() {
+				_, err := serviceBroker.Provision(context.TODO(), "new-instance", provisionDetails, true)
+				Expect(err).To(MatchError("error saving instance details to database: failed to store SI details. WARNING: this instance cannot be deprovisioned through cf. Contact your operator for cleanup"))
+			})
+		})
+
+		When("storage errors when storing provision parameters", func() {
```
pivotal-marcela-campo, line 299: I like that we can now mock these errors instead of having to have to produce a real error for testing as before!

pivotal-marcela-campo (changes requested):
There's a test that needs improvement. Otherwise it is great!

brokerapi/broker/provision_test.go
FelisiaM: There is scope to move the call of `request.DecodeOriginatingIdentityHeader` from the provision/deprovision and pass it to `service_definition.ProvisionVariables` function if we want to fully test it. But I found the setup for the old test confusing because its not what the real flow looks like

brokerapi/broker/provision_test.go
FelisiaM: Yeah this new test is aiming to keep the unit test boundaries and mock all interactions with other modules.

brokerapi/broker/deprovision_test.go
```diff
@@ -0,0 +1,281 @@
...
+			Expect(err).ToNot(HaveOccurred())
+
+			By("validating response")
+			Expect(response.IsAsync).To(BeFalse())
+			Expect(response.OperationData).To(BeEmpty())
+
+			By("validating call to deprovision")
+			Expect(fakeServiceProvider.DeprovisionCallCount()).To(Equal(1))
+			actualCtx, instanceID, actualDetails, actualsVars := fakeServiceProvider.DeprovisionArgsForCall(0)
+			Expect(actualCtx.Value(middlewares.OriginatingIdentityKey)).To(Equal(expectedHeader))
```
pivotal-marcela-campo, line 111: same comment about originatingIdentity header

pivotal-marcela-campo (approved):
Apart from the originatingIdentity tests that'll be added, looks good to me.

brokerapi/broker/provision_test.go
FelisiaM: Added the tests 🎉

pivotal-marcela-campo (approved):

PR #391, opened 2022-02-18
chore: Transfer more broker tests into ginkgo style

FelisiaM:
### Checklist:


brokerapi/broker/update.go
```diff
@@ -98,10 +96,12 @@ func (broker *ServiceBroker) Update(ctx context.Context, instanceID string, deta
 		return domain.UpdateServiceSpec{}, err
 	}
 
-	// save instance details
-	instance.PlanGUID = newInstanceDetails.PlanId
-	if err := broker.store.StoreServiceInstanceDetails(instance); err != nil {
-		return domain.UpdateServiceSpec{}, fmt.Errorf("error saving instance details to database: %s. WARNING: this instance cannot be deprovisioned through cf. Contact your operator for cleanup", err)
+	// save instance plan change
+	if instance.PlanGUID != details.PlanID {
```
jimbo459, line 100: We now only save instance details if the planID has changed. Is this the only thing which could have changed to trigger a need to save instance details?
I think this is OK as we do the separate save of provision details, but just wanted to check.

brokerapi/broker/update_test.go
```diff
@@ -0,0 +1,563 @@
...
+
+			It("should error", func() {
+				updateDetails.PlanID = newPlanID
+
+				_, err := serviceBroker.Update(context.TODO(), instanceID, updateDetails, true)
+				Expect(err).To(MatchError("error saving instance details to database: failed to store SI details. WARNING: this instance cannot be deprovisioned through cf. Contact your operator for cleanup"))
+			})
+		})
+
+		When("storage errors when storing provision parameters", func() {
```
blgm, line 552: Perhaps `Context()` would make more sense here?

PR #395, opened 2022-02-23
chore: Transferred last operation tests to ginkgo

FelisiaM:
### Checklist:

### DIFF
--- a/brokerapi/broker/last_instance_operation_test.go
+++ b/brokerapi/broker/last_instance_operation_test.go
@@ -1,0 +1,339 @@
+package broker_test
+
+import (
+	"errors"
+
+	"code.cloudfoundry.org/lager"
+	"github.com/cloudfoundry/cloud-service-broker/brokerapi/broker/brokerfakes"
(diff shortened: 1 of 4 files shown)


brokerapi/broker/last_instance_operation_test.go
```diff
@@ -0,0 +1,339 @@
...
+			})
+		})
+	})
+
+	Describe("operation in progress", func() {
+		BeforeEach(func() {
+			fakeServiceProvider.PollInstanceReturns(false, "operation in progress still", nil)
+		})
+
+		It("should complete provision", func() {
```
pivotal-marcela-campo, line 217: should not update SI, maybe? OR any other name that makes sense

pivotal-marcela-campo (changes requested):

brokerapi/broker/last_instance_operation_test.go
FelisiaM: Yeah you're right! Fixed it

PR #430, opened 2022-03-18
test: add broker unbind test

jimbo459:
### Checklist:

~* [ ] Have you added Draft Release Notes in `docs/draft-release-notes.md`?~


brokerapi/broker/unbind_test.go
```diff
@@ -0,0 +1,316 @@
...
+
+		It("should remove binding from database", func() {
+			expectedHeader := "cloudfoundry eyANCiAgInVzZXJfaWQiOiAiNjgzZWE3NDgtMzA5Mi00ZmY0LWI2NTYtMzljYWNjNGQ1MzYwIg0KfQ=="
+			newContext := context.WithValue(context.Background(), middlewares.OriginatingIdentityKey, expectedHeader)
+
+			response, err := serviceBroker.Unbind(newContext, instanceID, bindingID, unbindDetails, false)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("validating response")
+			Expect(response).To(BeZero())
```
FelisiaM, line 153: Technically correct, but it would be better to be nil

brokerapi/broker/unbind_test.go
```diff
@@ -0,0 +1,316 @@
...
+
+			By("validating provider unbind has been called")
+			Expect(fakeServiceProvider.UnbindCallCount()).To(Equal(1))
```
FelisiaM, line 156: Personal preference - but I would say here is a good opportunity to validate that unbind is being called with the expected args - instanceid and binding id)
Pro: For me this is important for calls to functions that accept a list of strings like this one, because it a the type of call where one can easily put the args in the wrong order since they are all just strings. 
Cons:
- one can say there will be integration tests that should fail if the args are in the wrong order (might be true but this codebase is weird)
- it makes refactoring that includes changing signatures more complex. Again in this codebase I don't see that happening anytime soon

brokerapi/broker/unbind_test.go
```diff
@@ -0,0 +1,316 @@
...
+			return fakeServiceProvider
+		}
+		planUpdatable := true
+
+		brokerConfig := &broker.BrokerConfig{
+			Registry: pkgBroker.BrokerRegistry{
+				"test-service": &pkgBroker.ServiceDefinition{
+					Id:   offeringID,
+					Name: "test-service",
+					Plans: []pkgBroker.ServicePlan{
```
FelisiaM, line 60: nitpick: There is quite a long setup to this service - I would suggest you remove things from the setup that do not change the behaviour of the test. E.g `ProvisionInputVariables` are not really used in unbind.
For me this helps make it clear what is setup intentionally

brokerapi/broker/unbind_test.go
```diff
@@ -0,0 +1,316 @@
...
+				PlanGUID:         planID,
+				SpaceGUID:        spaceID,
+				OrganizationGUID: orgID,
+				OperationType:    models.ProvisionOperationType,
+				OperationGUID:    "provision-operation-GUID",
+			}, nil)
+			fakeStorage.GetBindRequestDetailsReturns(json.RawMessage(`{"decrypted":{"foo":"bar"}}`), nil)
+		})
+
+		It("should remove binding from database", func() {
```
FelisiaM, line 145: There is another use case of unbinding when credhub is enabled - did you intentionally leave it out because it's not straightforward to test or was is an oversight?

brokerapi/broker/unbind_test.go
```diff
@@ -0,0 +1,316 @@
...
+
+			By("validating DeleteBindRequestDetails has been called")
+			Expect(fakeStorage.DeleteBindRequestDetailsCallCount()).To(Equal(1))
+			actualBindingId, actualInstanceID = fakeStorage.DeleteBindRequestDetailsArgsForCall(0)
+			Expect(actualBindingId).To(Equal(bindingID))
+			Expect(actualInstanceID).To(Equal(instanceID))
+		})
+
+	})
+
```
FelisiaM, line 174: One area that is not tested is the bind variables - this is a bit of a tricky one because not all of it can be purely unit tested, but I think it is worth doing a few tests inspired by the deprovision [tests](https://github.com/cloudfoundry/cloud-service-broker/blob/1ee046d61490ce42491c834b5c233a83753351d2/brokerapi/broker/deprovision_test.go#L132-L178)

FelisiaM (changes requested):
I added a few comments - some are just opinions (that you can ignore) and some are missing test cases.

Overall I think is so much improvement from what was already there.

Also, dot forget to remove the old test from `broker_test.go` as part of this PR

brokerapi/broker/unbind_test.go
jimbo459: I think this one does need to expect to be zero as a struct is returned with `zero` values ([see docs](https://go.dev/ref/spec#The_zero_value)). 
Asserting to equal nil fails as a struct is returned, just with zero values.

brokerapi/broker/unbind_test.go
jimbo459: Agreed - good catch

brokerapi/broker/unbind_test.go
jimbo459: I have added assertions that the credstore methods have been invoked and the correct amount of times. Faking the arguments they're called with is not really possible with the current implementation, but this feels like a good catch to ensure those methods are being called.

brokerapi/broker/unbind_test.go
jimbo459: Thanks - have added some tests around the vars

blgm (approved):

PR #444, opened 2022-04-05
test: switch to using pure go sqlite library

jimbo459:
Previously the sqlite driver required CGO to be enabled.
This change also uses sqlite as the DB if the hostname env variable starts with a '/'.
A timeout parameter was required in the connection string to avoid DB lock issues in multi-thread usage.

### Checklist:


db_service/setup_db.go
```diff
@@ -73,6 +74,12 @@ func SetupDb(logger lager.Logger) *gorm.DB {
 	dbType := viper.GetString(dbTypeProp)
 	var db *gorm.DB
 	var err error
+
+	if strings.HasPrefix(viper.GetString(dbHostProp), "/") {
```
tinygrasshopper, line 78: I dont think this if block makes sense, there is an env config `DB_TYPE` which does this for us

blgm (approved):
We noted that:
- the approach is a little hacky. For example even though `DB_TYPE` is set to `mysql`, it may actually use SQLite
- the new dependency is mentioned on the Gorm website, but all new dependencies have security implications. Currently this project uses quite a lot of dependencies of untested origin, so adding this one doesn't make a significant difference to the security posture.

These trade-offs are worth it to achieve the objective of better testing.

tinygrasshopper:
@jimbo459 I think this was achieved with alternate means, so are we ok to close to PR?

PR #453, opened 2022-04-13
feat(experimental): store *.pem files alongside Terraform state

blgm:
To support TLS for GCP PostgreSQL, it proved necessary to write
*.pem files for the Terraform provider to use to connect as a client.

This feature is experimental and will change, as the files that are
stored alongside the Terraform state should be configurable.

[#181059797](https://www.pivotaltracker.com/story/show/181059797)

### Checklist:

* [X] Have you added or updated tests to validate the changed functionality?
* [X] Have you added Draft Release Notes in `docs/draft-release-notes.md`?

### DIFF
--- a/cmd/tf.go
+++ b/cmd/tf.go
@@ -83,7 +83,8 @@ func init() {
 			}
 
 			if onlyState {
-				fmt.Printf("%s", string(ws.State))
+				fmt.Println(string(ws.State))
+				fmt.Printf("%+v\n", ws.AdditionalState)
 			} else {
 				fmt.Println(ws)
 			}
--- a/docs/draft-release-notes.md
+++ b/docs/draft-release-notes.md
@@ -2,6 +2,7 @@
 
 ### Features:
 - new "/info" endpoint for diagnostics reports broker version and uptime
+- "*.pem" files get stored alongside Terraform state - experimental and subject to change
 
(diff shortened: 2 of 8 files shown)


jimbo459:
@blgm It wasn't immediately clear to me from the naming `AdditionalState` that this state can only contain .pem files. 
Would it make sense to make this more concrete i.,e `PEMFiles` - given that the current implementation only allows type of file?

jimbo459 (changes requested):
See comment

blgm:
> @blgm It wasn't immediately clear to me from the naming `AdditionalState` that this state can only contain .pem files. Would it make sense to make this more concrete i.,e `PEMFiles` - given that the current implementation only allows type of file?

I think that would make sense if there was never a plan to change this. But because this data gets stored in the database, any change would require a database migration. So if today we were to have:
```go
PEMFiles map[string][]byte `json:"pem_files"`
```
And tomorrow want to change it to:
```go
AdditionalState map[string][]byte `json:"additional_state"`
```
Then we would need to apply this change to the whole database on startup. This is entirely possible, but work that's easily avoided by being a bit more generic.

jimbo459:
Makes a lot of sense @blgm - thanks for the clear explanation for the naming.

PR #455, opened 2022-04-14
feat: allow users to configure skip-verify when using custom certs

jimbo459:
[#179702988](https://www.pivotaltracker.com/story/show/179702988)

### Checklist:


blgm (approved):
It would be best to get rid of `InsecureSkipVerify` completely, but as there may be folks who rely on this behaviour, it makes sense to at least make it configurable.

PR #464, opened 2022-05-03
feat: terraform upgraded before unbind & delete

jimbo459:
### Checklist:


brokerapi/broker/unbind.go
```diff
@@ -65,7 +65,7 @@ func (broker *ServiceBroker) Unbind(ctx context.Context, instanceID, bindingID s
 	}
 
 	// remove binding from service provider
-	if err := serviceProvider.Unbind(ctx, instanceID, bindingID, vars); err != nil {
+	if err = serviceProvider.Unbind(ctx, instanceID, bindingID, vars); err != nil {
```
blgm, line 68: I think this was a little safer before as it reduced the scope of `err`, but I'm guessing this change was made because of an IDE warning.

integrationtest/fixtures/brokerpak-terraform-0.12/fake-service.yml
```diff
@@ -14,3 +14,6 @@ plans:
 provision:
   template_refs:
     main: fake-provision.tf
+bind:
+  template_refs:
+    main: fake-provision.tf
```
blgm, line 19: A little confusing for a fake binding to use `fake-provision`, but I'm guessing it doesn't matter

integrationtest/upgrade_terraform_before_delete_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+				By("deleting the service instance")
+				deleteBindResponse := testHelper.Client().Deprovision(serviceInstanceGUID, serviceOfferingGUID, servicePlanGUID, requestID())
+
+				Expect(deleteBindResponse.Error).NotTo(HaveOccurred())
+				Expect(deleteBindResponse.StatusCode).To(Equal(http.StatusAccepted))
+
+				Eventually(pollLastOperation(testHelper, serviceInstanceGUID), time.Minute*2, time.Second*10).Should(Equal(domain.Succeeded))
+
+				By("observing that the TF state file has been updated to the latest version before destroy")
```
blgm, line 84: This clearly works for now. But you could imagine a future where we delete the Terraform state when we delete a service instance. Is there another thing we could test for that is a behaviour, rather than an aspect of the implementation? Maybe there isn't.

integrationtest/upgrade_terraform_before_delete_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/cloudfoundry/cloud-service-broker/db_service/models"
+
+	"github.com/cloudfoundry/cloud-service-broker/integrationtest/helper"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gexec"
+)
+
+var _ = Describe("delete", func() {
```
blgm, line 22: Although the file name is descriptive, Ginkgo will just name this test as "integrationtest/delete".

pkg/providers/tf/job_runner.go
```diff
@@ -319,6 +320,12 @@ func (runner *TfJobRunner) Destroy(ctx context.Context, id string, templateVars
 	}
 
 	go func() {
+		err = runner.performTerraformUpgrade(ctx, workspace)
```
blgm, line 323: I'm guessing it wasn't easy to write a unit test for this. I know that there was a lot of refactoring of the job runner recently specifically in order to make it easier to test. Can you think of a refactoring that would make this easier to test directly?

blgm (approved):

brokerapi/broker/unbind.go
jimbo459: Yes, this was to satisfy the IDE...

jimbo459: Reverting as there are lots of other examples, so better to stay consistent rather than just fix this one.

integrationtest/fixtures/brokerpak-terraform-0.12/fake-service.yml
jimbo459: Yes - from a TF perspective there is no difference, it just expects some TF that it can invoke... 
The bind file would have identical content so felt a little superfluous.

integrationtest/upgrade_terraform_before_delete_test.go
jimbo459: This specific step is just reading the output of Terraform commands, rather than querying state. I will amend the `By` as it is misleading here. 

You are right that in the second test we query the state file - this step is potentially not actually needed, I was just a little anxious to ensure that the failure to delete was actually due to the TF binary being > version in the statefile and couldn't see another way of testing this.

pkg/providers/tf/job_runner.go
```diff
@@ -319,6 +320,12 @@ func (runner *TfJobRunner) Destroy(ctx context.Context, id string, templateVars
...
+		if err != nil {
+			runner.operationFinished(err, workspace, deployment)
+			return
+		}
+
 		err = runner.DefaultInvoker().Destroy(ctx, workspace)
 		runner.operationFinished(err, workspace, deployment)
```
zucchinidev, line 330: thoughts: it seems that `runner.operationFinished` method is always called, maybe it is a candidate to perform the operation in a defer

zucchinidev, line 330: thoughts: it seems that `runner.operationFinished` method is always called, maybe it is a candidate to perform the operation in a defer

zucchinidev: It makes totally sense. We would lose readability and obscure the solution. There are more uses of that approach in the same file and we would end up with different approaches. Thanks!

pkg/providers/tf/job_runner.go
jimbo459: This is a fair challenge - this function is actually unit tested in job_runner_test, but as part of the update flow. 
We should add tests for the Destroy method...

PR #465, opened 2022-05-04
chore: fix some initialisms

blgm:
https://github.com/golang/go/wiki/CodeReviewComments#initialisms
Go guidelines (and statickcheck rule ST1003) recommend using ID, JSON,
etc... rather than Id, Json, etc...

### Checklist:


pkg/providers/tf/provider.go
```diff
@@ -174,84 +174,84 @@ func (provider *terraformProvider) importCreate(ctx context.Context, vars *varco
...
 
 	workspace, err := workspace.NewWorkspace(vars.ToMap(), action.Template, action.Templates, []workspace.ParameterMapping{}, []string{}, []workspace.ParameterMapping{})
 	if err != nil {
-		return tfId, fmt.Errorf("error creating workspace: %w", err)
+		return tfID, fmt.Errorf("error creating workspace: %w", err)
 	}
 
 	// if err = workspace.Validate(); err != nil {
-	// 	return tfId, err
+	// 	return tfID, err
```
zucchinidev, line 207: **thoughts**: perhaps it is worth deleting this code fragment

pkg/providers/tf/provider.go
```diff
@@ -174,84 +174,84 @@ func (provider *terraformProvider) importCreate(ctx context.Context, vars *varco
...
 
 // Deprovision performs a terraform destroy on the instance.
-func (provider *terraformProvider) Deprovision(ctx context.Context, instanceGUID string, details domain.DeprovisionDetails, vc *varcontext.VarContext) (operationId *string, err error) {
+func (provider *terraformProvider) Deprovision(ctx context.Context, instanceGUID string, details domain.DeprovisionDetails, vc *varcontext.VarContext) (operationID *string, err error) {
 	provider.logger.Debug("terraform-deprovision", correlation.ID(ctx), lager.Data{
 		"instance": instanceGUID,
 	})
 
-	tfId := generateTfId(instanceGUID, "")
+	tfID := generateTfId(instanceGUID, "")
```
zucchinidev, line 244: **question**: following Go idioms, what do you think about change the name of this method to `generateTfID`

pkg/providers/tf/provider.go
```diff
@@ -35,7 +35,7 @@ import (
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
 //counterfeiter:generate . JobRunner
 type JobRunner interface {
-	StageJob(jobId string, workspace *workspace.TerraformWorkspace) error
+	StageJob(jobID string, workspace *workspace.TerraformWorkspace) error
 	Import(ctx context.Context, id string, importResources []ImportResource) error
 	Create(ctx context.Context, id string) error
 	Update(ctx context.Context, id string, templateVars map[string]interface{}) error
```
zucchinidev, line 41: **question**: Do you think we should change the parameter `id` to `ID`?

zucchinidev:
@blgm Thank you very much for continuing to persevere in the conventions of Go. Great job!. I added a couple of questions, but for everything else correct.

pkg/providers/tf/provider.go
blgm: No. Starting a variable name with a capital letter suggests that it should be exported.

pkg/providers/tf/provider.go
blgm: This PR only fixes _some_ initialisms. If you enable rule ST1003 you'll see that there are many more.

pkg/providers/tf/provider.go
blgm: Maybe. This story was about fixing some initialisms, so I think that's outside the scope of this PR.

pkg/providers/tf/provider.go
zucchinidev: Understood!  Thank you very much for the answer.

PR #475, opened 2022-05-10
chore: fix some more initialisms

blgm:
Follows on from commits bdd1de9, f568553, and 459a72f

https://github.com/golang/go/wiki/CodeReviewComments#initialisms
Go guidelines (and statickcheck rule ST1003) recommend using ID, JSON,
etc... rather than Id, Json, etc...

### Checklist:

### DIFF
--- a/db_service/models/historical_db.go
+++ b/db_service/models/historical_db.go
@@ -34,9 +34,9 @@ type ServiceBindingCredentialsV1 struct {
 
 	OtherDetails string `gorm:"type:text"`
 
-	ServiceId         string
-	ServiceInstanceId string
-	BindingId         string
+	ServiceID         string
+	ServiceInstanceID string
+	BindingID         string
 }
 
 // TableName returns a consistent table name for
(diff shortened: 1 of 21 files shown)


zucchinidev:
Great job @blgm, thank you very much!

db_service/models/historical_db.go
```diff
@@ -135,24 +135,24 @@ type ServiceInstanceDetailsV3 struct {
...
 	Location     string
-	Url          string
+	URL          string
 	OtherDetails []byte `gorm:"type:blob"`
 
-	ServiceId        string
-	PlanId           string
-	SpaceGuid        string
-	OrganizationGuid string
+	ServiceID        string
```
pivotal-marcela-campo, line 141: Just wondering how does GORM treats this change. I am asuming field names will continue to be the same and `ServiceID` will be changed to `service_id` in the database as it was before. Is that the case?

pivotal-marcela-campo (commented):
Looks ok, just have a question about the change in field names in models.

db_service/models/historical_db.go
blgm: That's a good question. I looked at the GORM tests and think it should be ok: https://github.com/go-gorm/gorm/blob/373bcf7aca01ef76c8ba5c3bc1ff191b020afc7b/schema/naming_test.go#L10-L26

pivotal-marcela-campo (approved):

PR #482, opened 2022-05-12
chore: refactor, removed async supported from the service provider interface

tinygrasshopper:
- We dont really have any providers that dont support async
- Reduces the number of methods on the service provider interface

Co-authored-by: Marcela Campo <mcampo@vmware.com>

### Checklist:

~~* [ ] Have you added Draft Release Notes in `docs/draft-release-notes.md`?~~


brokerapi/broker/deprovision.go
```diff
@@ -48,8 +48,7 @@ func (broker *ServiceBroker) Deprovision(ctx context.Context, instanceID string,
 		return response, err
 	}
 
-	// if async deprovisioning isn't allowed but this service needs it, throw an error
-	if serviceProvider.DeprovisionsAsync() && !clientSupportsAsync {
+	if !clientSupportsAsync {
```
blgm, line 51: Potentially this could be on the first line of this method.

blgm (approved):

brokerapi/broker/deprovision.go
zucchinidev, line 51: thoughts: it seems the `Deprovision` method is only called from `handlers/deprovision.go:51`. We inject the value of the `asyncAllowed` parameter depending on the value of the `accepts_incomplete` form parameter. It may be worth doing the validation in the top layer and modifying the `Deprovision` public interface. Obviously, this change is out of the scope of this PR.

brokerapi/broker/provision.go
```diff
@@ -61,8 +61,7 @@ func (broker *ServiceBroker) Provision(ctx context.Context, instanceID string, d
 	}
 
 	// verify async provisioning is allowed if it is required
-	shouldProvisionAsync := serviceProvider.ProvisionsAsync()
-	if shouldProvisionAsync && !clientSupportsAsync {
+	if !clientSupportsAsync {
```
zucchinidev, line 64: same comment as the previous

brokerapi/broker/provision.go
```diff
@@ -104,5 +103,5 @@ func (broker *ServiceBroker) Provision(ctx context.Context, instanceID string, d
 		return domain.ProvisionedServiceSpec{}, fmt.Errorf("error saving provision request details to database: %s. Services relying on async provisioning will not be able to complete provisioning", err)
 	}
 
-	return domain.ProvisionedServiceSpec{IsAsync: shouldProvisionAsync, DashboardURL: "", OperationData: instanceDetails.OperationGUID}, nil
+	return domain.ProvisionedServiceSpec{IsAsync: true, DashboardURL: "", OperationData: instanceDetails.OperationGUID}, nil
```
zucchinidev, line 106: **thoughts**: this property should probably disappear

zucchinidev (approved):

brokerapi/broker/deprovision.go
pivotal-marcela-campo: @zucchinidev handlers/deprovision.go:51 is part of the [brokerapi](https://github.com/pivotal-cf/brokerapi) project which defines a common implementation of the OSBAPI spec to ease the development of OSBAPI compliant brokers. As such CSB is not the only broker that is implemeted based on it. The check we are doing here is basically because our broker does not support synchronous provisioning at all. So if the client is expecting the provision to be completed synchronously we just return an error.

brokerapi/broker/provision.go
pivotal-marcela-campo: Same as the previous comments, it is part of the serviceapi contract and the returned response will depend on this property. For e.g [here for deprovision](https://github.com/pivotal-cf/brokerapi/blob/813290b2c03139287314af01af39deb355931e30/handlers/deprovision.go#L66). 
It is food for thought though if the `brokerapi` could make this decision based on something else, like the presence of operation data. It is definitely out of scope from this PR and would need a PR into the `serviceapi` project.

brokerapi/broker/deprovision.go
zucchinidev: Fantastic @pivotal-marcela-campo. Thank you very much for the explanation . Definitely, I need to study the public API

PR #517, opened 2022-05-24
feat: terraform upgrades no longer performed on update/delete

jimbo459:
- This is in preparation of the new workflow where Terraform will
  only by upgraded as part of the maintenance upgrade flow

[#182209076](https://www.pivotaltracker.com/story/show/182209076)

### Checklist:


FelisiaM:
I think the behaviour change should be included in the docs release notes as well. Last release notes say that delete/unbind can do upgrade as well

pkg/providers/tf/update.go
```diff
@@ -44,7 +41,7 @@ func (provider *TerraformProvider) Update(ctx context.Context, provisionContext
 	}
 
 	go func() {
-		err = provider.performTerraformUpgrade(ctx, workspace)
```
FelisiaM, line 47: is this an early optimisation? Just thinking that there is a store to split the upgrade flow and that will need to implement the terraform upgrade. By removing it it means we have to rewrite the same code and tests again instead of just moving them around

pkg/providers/tf/provider.go
```diff
@@ -160,6 +161,17 @@ func (provider *TerraformProvider) Wait(ctx context.Context, id string) error {
 	}
 }
 
+func (provider *TerraformProvider) checkTerraformVersion(ctx context.Context, workspace workspace.Workspace) error {
```
FelisiaM, line 164: context is no longer being used or needed here

pkg/providers/tf/provider.go
```diff
@@ -160,6 +161,17 @@ func (provider *TerraformProvider) Wait(ctx context.Context, id string) error {
...
+	currentTfVersion, err := workspace.StateVersion()
+	if err != nil {
+		return err
+	}
+	if currentTfVersion.LessThan(provider.tfBinContext.DefaultTfVersion) {
+		return errors.New("apply attempted with a newer version of terraform than the state")
```
FelisiaM, line 170: Is this essentially implementing part of the next story about blocking all interactions with the delete/unbind https://www.pivotaltracker.com/story/show/182208954 ?

pkg/providers/tf/provider.go
```diff
@@ -132,7 +133,7 @@ func (provider *TerraformProvider) destroy(ctx context.Context, deploymentID str
 	}
 
 	go func() {
-		err = provider.performTerraformUpgrade(ctx, workspace)
+		err = provider.checkTerraformVersion(ctx, workspace)
```
FelisiaM, line 135: I think this check should not be be in the goroutine. The check should be done as soon as we get the workspace and fail fast as part of the initial response, not as a last operation failure.
There is another question in my mind about what error should be returned. I think the message should mention a tip to run upgrade-service, but possible also respond with a `MaintenanceInfoConflict` error

pkg/providers/tf/update.go
jimbo459: The spike has all the code / tests. We essentially need to remove the upgrade from update in its current position before starting on the upgrade flow. 
This felt like a clean way to reset us back to a clean state where there is no upgrade performed, but the code is archived to be used in the actual implementation.

FelisiaM (changes requested):
Left some comments

pkg/providers/tf/provider.go
jimbo459: I was going to strip all this out, but we have now exposed the ability to add multiple Terraform Binaries to a Brokerpak. It felt a little dangerous to remove this step, given that it already existed, but just remove the upgrade logic. 
You're right that story will essentially result in the same behaviour, but fail at an earlier point in the flow.

pkg/providers/tf/provider.go
jimbo459: Yeah - in my mind this would all be done as part of https://www.pivotaltracker.com/story/show/182208954. This chore was simply stripping out the upgrade behaviour and leaving the rest of the current logic untouched, with a view that it will all be solved in that story.

pkg/providers/tf/update.go
FelisiaM: Being realistic,  maybe it's enough to just added a link to this change in the next story so its easy to find.

FelisiaM: Maybe the part that might not be clear is that update path would not be exercised anymore so there is no need for this check here at all. 
Update and upgrade go through the same endpoint and also maintenance info is always passed by CF.

jimbo459:
Closing as we will play story https://www.pivotaltracker.com/story/show/182208954 which will cover this functionality.

PR #520, opened 2022-05-26
feat: Use maintenance info to trigger upgrades

FelisiaM:
[#182208628](https://www.pivotaltracker.com/story/show/182208628)

### Checklist:


brokerapi/broker/broker.go
```diff
@@ -120,3 +125,9 @@ func generateTFInstanceID(instanceID string) string {
 func generateTFBindingID(instanceID, bindingID string) string {
 	return "tf:" + instanceID + ":" + bindingID
 }
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
+//counterfeiter:generate . Decider
+type Decider interface {
```
blgm, line 131: Did you consider not making this an interface? Sometimes it makes sense to make something an interface for testing reasons, and sometimes (in my opinion) it introduces more noise than benefit.

brokerapi/broker/decider/decider.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+type Decider struct{}
+
+type Operation int
+
+const (
+	Update Operation = iota
+	Upgrade
+	Failed
```
blgm, line 22: I would make `Failed` the zero-value so that an uninitialised `Update` is `Failed` rather than `Update`.

brokerapi/broker/decider/decider.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+var errInstanceMustBeUpgradedFirst = apiresponses.NewFailureResponseBuilder(
```
blgm, line 25: I slightly prefer making these functions so that no-one can accidentally change the value later. For example, if the line `errInstanceMustBeUpgradedFirst=errMaintenanceInfoNilInTheRequest` were to be added somewhere, it would be very hard to debug.

brokerapi/broker/decider/decider.go
```diff
@@ -0,0 +1,143 @@
...
+
+	var params map[string]interface{}
+	if err := json.Unmarshal(details.RawParameters, &params); err != nil {
+		return false
+	}
+	return len(params) == 0
+}
+
+func requestMaintenanceInfoValuesDiffer(details domain.UpdateDetails) bool {
+	if details.MaintenanceInfo == nil && details.PreviousValues.MaintenanceInfo != nil {
```
blgm, line 70: Alternatively:
```go
switch {
  case  details.MaintenanceInfo == nil && details.PreviousValues.MaintenanceInfo != nil:
    return true
  case details.MaintenanceInfo != nil && details.PreviousValues.MaintenanceInfo == nil:
    return true
  case details.MaintenanceInfo == nil && details.PreviousValues.MaintenanceInfo == nil:
    return false
  default:
     return !details.MaintenanceInfo.Equals(*details.PreviousValues.MaintenanceInfo)
}
```
Personally I think this makes it clearer that only one can be true, but folks have differing opinions.

brokerapi/broker/decider/decider.go
```diff
@@ -0,0 +1,143 @@
...
+	return !details.MaintenanceInfo.Equals(*details.PreviousValues.MaintenanceInfo)
+}
+
+func validateMaintenanceInfo(service *broker.ServiceDefinition, planID string, maintenanceInfo *domain.MaintenanceInfo) error {
+	planMaintenanceInfo, err := getMaintenanceInfoForPlan(service, planID)
+	if err != nil {
+		return err
+	}
+
+	if maintenanceInfoConflict(maintenanceInfo, planMaintenanceInfo) {
```
blgm, line 91: Maybe `maintenanceInfoConflict` could return the correct error?

brokerapi/broker/decider/decider.go
```diff
@@ -0,0 +1,143 @@
...
+				}, nil
+			}
+			return nil, nil
+		}
+	}
+
+	return nil, fmt.Errorf("plan %s does not exist", id)
+}
+
+func maintenanceInfoConflict(a, b *domain.MaintenanceInfo) bool {
```
blgm, line 133: This looks like the same algorithm as `requestMaintenanceInfoValuesDiffer()`. Is there scope to reduce duplication?

brokerapi/broker/update.go
```diff
@@ -19,7 +23,7 @@ var ErrNonUpdatableParameter = apiresponses.NewFailureResponse(errors.New("attem
 
 // Update a service instance plan.
 // This functionality is not implemented and will return an error indicating that plan changes are not supported.
-func (broker *ServiceBroker) Update(ctx context.Context, instanceID string, details domain.UpdateDetails, asyncAllowed bool) (response domain.UpdateServiceSpec, err error) {
+func (broker *ServiceBroker) Update(ctx context.Context, instanceID string, details domain.UpdateDetails, asyncAllowed bool) (domain.UpdateServiceSpec, error) {
```
blgm, line 26: Nice improvement. The previous use of `response` was a bug waiting to happen.

brokerapi/broker/update.go
```diff
@@ -53,49 +57,68 @@ func (broker *ServiceBroker) Update(ctx context.Context, instanceID string, deta
...
 	}
 
 	vars, err := serviceDefinition.UpdateVariables(instanceID, parsedDetails, mergedDetails, *plan, request.DecodeOriginatingIdentityHeader(ctx))
 	if err != nil {
-		return response, err
+		return domain.UpdateServiceSpec{}, err
+	}
+
+	operation, err := broker.decider.DecideOperation(serviceDefinition, details)
+	if err != nil {
```
blgm, line 97: Alternatively:
```go
switch {
  case err != nil:
    return domain.UpdateServiceSpec{}, fmt.Errorf("error deciding update path: %w", err)
  case operation == decider.Upgrade:
    return broker.doUpgrade(ctx, serviceProvider, vars)
  default:
    return broker.doUpdate(ctx, serviceProvider, instance, vars, parsedDetails, mergedDetails)
}
```
Personally I think this makes it very clear that it's a thee way branch.

Also, given we don't actually check for `decider.Failure` and `decider.Update`, would it be simpler to use a boolean `isUpgrade` rather than introducing a new type?

integrationtest/helper/operations.go
```diff
@@ -40,7 +40,15 @@ func (h *TestHelper) Provision(serviceOfferingGUID, servicePlanGUID string, para
...
 func (h *TestHelper) UpdateService(s ServiceInstance, params ...any) {
 	const offset = 1
-	updateResponse := h.Client().Update(s.GUID, s.ServiceOfferingGUID, s.ServicePlanGUID, uuid.New(), toJSONRawMessage(params, 2))
+	updateResponse := h.Client().Update(s.GUID, s.ServiceOfferingGUID, s.ServicePlanGUID, uuid.New(), toJSONRawMessage(params, 2), domain.PreviousValues{}, nil)
+	gomega.Expect(updateResponse.Error).WithOffset(offset).NotTo(gomega.HaveOccurred())
+	gomega.Expect(updateResponse.StatusCode).WithOffset(offset).To(gomega.Equal(http.StatusAccepted), string(updateResponse.ResponseBody))
+	gomega.Expect(h.LastOperationFinalState(s.GUID, 1)).WithOffset(offset).To(gomega.Equal(domain.Succeeded))
+}
+
+func (h *TestHelper) UpdateServiceMI(s ServiceInstance, previousValues domain.PreviousValues, newMaintenanceInfo domain.MaintenanceInfo, params ...any) {
```
blgm, line 49: Could we call this `UpgradeService`?

pkg/broker/service_definition.go
```diff
@@ -502,13 +502,13 @@ func buildAndValidate(builder *varcontext.ContextBuilder, vars []BrokerVariable)
 	return vc, nil
 }
 
-func (svc *ServiceDefinition) AllowedUpdate(params map[string]interface{}) (bool, error) {
+func (svc *ServiceDefinition) AllowedUpdate(params map[string]interface{}) bool {
```
blgm, line 505: Thanks for cleaning this up

blgm (approved):

brokerapi/broker/broker.go
FelisiaM: Yes, that's how it started, however, the number of use cases that should be exercised would make the broker tests very large. This method itself has 24 tests as it stands for all permutations of previous MI, new MI and catalog MI.

brokerapi/broker/decider/decider_test.go
```diff
@@ -0,0 +1,435 @@
...
+							}
+
+							operation, err := decider.Decider{}.DecideOperation(serviceDefinition, details)
+							Expect(err).To(MatchError("maintenance info defined in broker service catalog, but not passed in request"))
+							Expect(operation).To(Equal(decider.Failed))
+						})
+					})
+				})
+
+				When("previous maintenance_info is present in the request", func() {
```
FelisiaM, line 187: Considering changing this to return "upgrade"

PR #521, opened 2022-05-26
fix: broker fails to delete service when create in progress

zucchinidev:
We want to avoid an error found when trying to delete a service when its creation is in progress as the service is not consolidated and the status of `Terraform` is not generated.

We also avoided an error when reading the `Terraform` state if that state was not previously generated, which generated an error when performing the `Unmarshalling` operation.

[#182125560](https://www.pivotaltracker.com/story/show/182125560)

### Checklist:

### DIFF
--- a/pkg/providers/tf/deprovision.go
+++ b/pkg/providers/tf/deprovision.go
@@ -12,7 +12,7 @@ import (
 )
 
 // Deprovision performs a terraform destroy on the instance.
-func (provider *TerraformProvider) Deprovision(ctx context.Context, instanceGUID string, details domain.DeprovisionDetails, vc *varcontext.VarContext) (operationID *string, err error) {
+func (provider *TerraformProvider) Deprovision(ctx context.Context, instanceGUID string, _ domain.DeprovisionDetails, vc *varcontext.VarContext) (*string, error) {
 	provider.logger.Debug("terraform-deprovision", correlation.ID(ctx), lager.Data{
 		"instance": instanceGUID,
 	})
--- a/pkg/providers/tf/deprovision_test.go
+++ b/pkg/providers/tf/deprovision_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/cloudfoundry/cloud-service-broker/dbservice/models"
 	"github.com/hashicorp/go-version"
 
 	"github.com/cloudfoundry/cloud-service-broker/internal/storage"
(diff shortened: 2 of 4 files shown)


pkg/providers/tf/deprovision_test.go
```diff
@@ -90,7 +112,7 @@ var _ = Describe("Deprovision", func() {
 		Expect(fakeDeploymentManager.MarkOperationStartedCallCount()).To(Equal(1))
 		actualDeployment, actualOperationType := fakeDeploymentManager.MarkOperationStartedArgsForCall(0)
 		Expect(actualDeployment).To(Equal(&deployment))
-		Expect(actualOperationType).To(Equal("deprovision"))
+		Expect(actualOperationType).To(Equal(models.DeprovisionOperationType))
```
FelisiaM, line 115: This would be a good improvement if we make sure the `models.DeprovisionOperationType` is also unit tested to prevent any unintended changes.

pkg/providers/tf/provider.go
```diff
@@ -112,6 +113,10 @@ func (provider *TerraformProvider) destroy(ctx context.Context, deploymentID str
 		return err
 	}
 
+	if deployment.LastOperationType == models.ProvisionOperationType && deployment.LastOperationState == InProgress {
```
FelisiaM, line 116: This method is used by both deprovision and unbind operations. While it is unlikely to be requesting an unbind while provision is in progress, adding a check on the current operation type being deprovision would make this more intentional.

pkg/providers/tf/provider.go
```diff
@@ -112,6 +113,10 @@ func (provider *TerraformProvider) destroy(ctx context.Context, deploymentID str
...
+		return fmt.Errorf("destroy operation not allowed - reason: provision in progress - tf ID: %s", deploymentID)
+	}
```
FelisiaM, line 118: I like you've added so much info on the error. Maybe a style more in line with the rest of the errors in the codebase would be : "destroy operation not allowed while provision is in progress"

This error would be exposed to the developer, who does not have a notion of what the tfID is, nor do we log it elsewhere. In reality, the TF ID will be something that looks like "tf:service-instance-GUID:". So if they needed to ask the operator to further investigate they already have the service instance guid and it can be tied to the request.

FelisiaM (changes requested):

FelisiaM:
Since this is a bug fix there should also be release notes added in `docs/draft-release-notes.md`

pkg/providers/tf/deprovision.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 // Deprovision performs a terraform destroy on the instance.
-func (provider *TerraformProvider) Deprovision(ctx context.Context, instanceGUID string, details domain.DeprovisionDetails, vc *varcontext.VarContext) (operationID *string, err error) {
+func (provider *TerraformProvider) Deprovision(ctx context.Context, instanceGUID string, _ domain.DeprovisionDetails, vc *varcontext.VarContext) (*string, error) {
```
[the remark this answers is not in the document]
zucchinidev: Thank you very much @FelisiaM  for your comment. I followed your recommendations, which has involved changes in different tests.

pkg/providers/tf/deprovision_test.go
```diff
@@ -21,45 +22,58 @@ import (
...
+			    }`,
+			Definitions: map[string]string{"variables": `
+				    variable name {type = "string"}
+					variable storage_class {type = "string"}
+				`},
+		}
+		moduleInstance := workspace.ModuleInstance{
+			ModuleName:   "cloud_storage",
+			InstanceName: "fake-instance-name",
+		}
```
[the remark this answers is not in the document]
zucchinidev: Done!!

pkg/providers/tf/deprovision_test.go
zucchinidev: Thank you @FelisiaM!!
I have decided not to add more changes and use the text string to no longer extend the scope of the PR.

I will create another unit test at another time associated with this change and so do not mix several concepts in a same pull request. Do you agree or do you want me to do it now?

pkg/providers/tf/deprovision_test.go
```diff
@@ -135,14 +178,76 @@ var _ = Describe("Deprovision", func() {
...
+		_, err := provider.Deprovision(context.TODO(), instanceGUID, domain.DeprovisionDetails{}, deprovisionContext)
+
+		Expect(err).To(
+			MatchError("destroy operation not allowed - reason: provision in progress - tf ID: " + expectedTFID),
+		)
+		Expect(fakeDeploymentManager.MarkOperationStartedCallCount()).To(Equal(0))
+	})
 })
+
+func newTfState() []byte {
```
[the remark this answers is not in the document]
zucchinidev: Done!!!

pkg/providers/tf/provider.go
zucchinidev: Thank you for your recommendations and explanations. Now I understand much more!! I have put the text you suggest

pkg/providers/tf/provider.go
zucchinidev: I have added the logic that you told me, but I have abstracted it to a method to improve readability

FelisiaM (approved):

PR #522, opened 2022-05-27
feat: maintenance info set to default tf version for user plans

jimbo459:
- Previously plans set in the brokerpak service definition had
  the maintenance info value populated, but plans provided in
  the csb config or as env vars did not

[#182296561](https://www.pivotaltracker.com/story/show/182296561)

### Checklist:


pkg/broker/registry.go
```diff
@@ -45,14 +45,14 @@ type BrokerRegistry map[string]*ServiceDefinition
...
-func (brokerRegistry BrokerRegistry) Register(service *ServiceDefinition) error {
+func (brokerRegistry BrokerRegistry) Register(service *ServiceDefinition, tfVersion string) error {
 	name := service.Name
 
 	if _, ok := brokerRegistry[name]; ok {
 		return fmt.Errorf("tried to register multiple instances of: %q", name)
 	}
 
-	userPlans, err := service.UserDefinedPlans()
+	userPlans, err := service.UserDefinedPlans(tfVersion)
```
FelisiaM, line 55: What do you think about the signature of this to be : UserDefinedPlans(maintenanceInfo &domain.MaintenanceInfo)
I am actually thinking that `toDefinitions(services, pak, tfBinariesContext)` also should take maintenanceInfo instead of tfcontext. 

That decouples the content of maintenance info from the assigning of it to the plans

jimbo459: `toDefinitions()` still needs the tfBinariesContext for supplying values to the executor factory, so I don't think we can remove that arg. 
I agree there is benefit in defining `maintenanceInfo` only once, if we were to go with your above suggestion we could pass it into `toDefinitions()` in addition to the tfBinariesContext.

The one thing I do like about the current pattern is we're only passing strings down through the methods, so there is no coupling of domain types through all the different layers. 
In an ideal world I would refactor this whole flow so plans are all parsed and created at once, rather than the two different branches for creating plans which were defined in the brokerpak and plans in config/env vars.

jimbo459: Moved to only using one maintenance info object as per suggestion.

PR #525, opened 2022-06-02
test: set initial state of Terraform

zucchinidev:
We need to set the initial state of Terraform to establish the Terraform
version.

[#182362436](https://www.pivotaltracker.com/story/show/182362436)

### Checklist:


jimbo459:
Hey @zucchinidev - I like the logic here of ensuring that the state is initialised. However, there may be legitimate test cases where we don't want the state to be instantiated when the terraform mock is returned. i.e before provisioning the state should be `nil`, whereas this change would instantiate the state before provision.

Given that we can resolve this issue by a change in the brokerpak, for now I think we can close this PR, but it raises a very good question around the interface of our mock. 
i.e should we follow the pattern found in our counterfeiter mocks where we would have `mockTerraform.TFStateReturns()`. In our unit tests, we need to define that return value if asserting on it, so I'd be inclined to follow a similar pattern here, rather than instantiate an object behind the scenes.

jimbo459:
Hi @zucchinidev - on reflection, the benefit of having the state instantiated when the TF mock is returned, outweighs the cost of having to have implicit knowledge in the test setup. 
I have made the change you suggested, but didn't add the TFInitState wrapper as that felt an unnecessary step given our current requirement. Happy to chat this through 🙂

PR #541, opened 2022-06-15
feat: allow upgrade of Terraform modules

blgm:
Terraform modules are a feature in Terraform. Most of our work
on the Terraform upgrade path has specified Terraform via `template_refs`
which results in not using Terraform modules. However if Terraform is
specified via `template_ref` then a single Terraform module is created.
The outputs from this module used not to be marked as "sensitive" which
meant that Terraform would fail if an output referred to a sensitive
value, which meant that we could not upgrade past Terraform 0.14. This
commit marks all the module outputs as sensitive. This is ignored by
earlier versions of Terraform, and allows for services using Terraform
modules to be upgraded to the latest Terraform version.

[#182413672](https://www.pivotaltracker.com/story/show/182413672)

### Checklist:

### DIFF
--- a/docs/draft-release-notes.md
+++ b/docs/draft-release-notes.md
@@ -10,6 +10,7 @@
     - Update endpoint can perform upgrades when the correct maintenance info information is passed and no other changes
       are requested. This will upgrade the instance and all related bindings.
     - Update, bind, unbind and delete operations are blocked if an upgrade has not happened first.
+    - Terraform modules can also be upgraded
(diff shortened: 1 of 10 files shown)


jimbo459:
Hey @blgm - have we checked this with a subsumed service? In my mind I'm concerned that the way we currently read values for subsume requires them not to be marked as sensitive?

blgm:
I've run the subsume tests with this commit and they do pass, so I think they are compatible.

PR #550, opened 2022-06-22
fix: ability to update after an upgrade

blgm:
Once a service instance has been upgraded, Cloud Foundry will put the
service instance's current Maintenance Info into the "previous values"
section of an update request. There's a relatively subtle
differentiation between:
- Downgrading a service instance (where it currently has a maintenance
info version and this is being removed), which is an Upgrade
- Updating a service instance (where there's a previous Maintenance Info
value, and no new Maintenance Info value) where there are other updates

The previous algorithm was hard to reason about, so it has been
redeveloped by adding many more table tests to solidify the behaviour.

As a general rule, Cloud Foundry provides a "previous values" block
which has all the current values, and only provides a new value when it
changes.

[#182522081](https://www.pivotaltracker.com/story/show/182522081)

### Checklist:

### DIFF
--- a/brokerapi/broker/decider/decider.go
+++ b/brokerapi/broker/decider/decider.go
@@ -23,7 +23,7 @@ const upgradeBeforeUpdateError = "service instance needs to be upgraded before u
 // of plan/parameters, or an Upgrade of Maintenance Info version. Is it an error to perform both at the
 // same time. CloudFoundry provides only the fields it intends to change in the request body,
 // but provides previous values for all fields. Where there is ambiguity, we rely on this CloudFoundry behavior.
-func DecideOperation(planMaintenanceInfoVersion *version.Version, details paramparser.UpdateDetails) (Operation, error) {
+func DecideOperation(serviceMaintenanceInfoVersion *version.Version, details paramparser.UpdateDetails) (Operation, error) {
 	requestHasMI := details.MaintenanceInfoVersion != nil
 	requestHasPreviousMI := details.PreviousMaintenanceInfoVersion != nil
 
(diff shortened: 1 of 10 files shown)


brokerapi/broker/decider/decider.go
```diff
@@ -19,61 +19,46 @@ const (
...
+	// Valid options would be:
+	// - update to plan/parameters with previous MI the same as plan MI
+	// - upgrade to the no-MI plan with no updates of plan/parameters
+	invalidUpdateAndMIRemoval := requestHasUpdate && !requestHasMI && requestHasPreviousMI && planMaintenanceInfoVersion == nil
+
+	switch {
+	case requestHasMI && planMaintenanceInfoVersion == nil:
+		// new MI is specified in request, but plan does not have MI
+		return Failed, apiresponses.ErrMaintenanceInfoNilConflict
+	case requestHasMI && !planMaintenanceInfoVersion.Equal(details.MaintenanceInfoVersion):
```
pivotal-marcela-campo, line 49: Can this maybe be merged with the previous case, maybe inverting the Equals as we know the request does have MI?

pivotal-marcela-campo (approved):
It looks good to me. Just left a couple of comments on things I found difficult to reason about.

brokerapi/broker/decider/decider_test.go
```diff
@@ -62,19 +59,48 @@ var _ = DescribeTable(
...
+	Entry(nil, "plan at v1;    MI none->v1; plan change;    no params", decider.Failed, "service instance needs to be upgraded before updating"),
+	Entry(nil, "plan at v1;    MI none->v1; plan unchanged; params", decider.Failed, "service instance needs to be upgraded before updating"),
+	Entry(nil, "plan at v1;    MI none->v1; plan change;    params", decider.Failed, "service instance needs to be upgraded before updating"),
+
+	// Attempted upgrades where the requested MI does not match the plan MI
+	Entry(nil, "plan at v1;    MI none->v2; plan unchanged; no params", decider.Failed, apiresponses.ErrMaintenanceInfoConflict),
+	Entry(nil, "no MI in plan; MI none->v1; plan unchanged; no params", decider.Failed, apiresponses.ErrMaintenanceInfoNilConflict),
+
+	// Edge case: updates where MI is held constant
+	// With CloudFoundry, new MI is only specified if it's different to previous MI, so we do not see this.
```
FelisiaM, line 94: That explenation is not correct. CAPI sends the new MI version if the client (cf cli or other) has passed it in the request. So in reality it is send when the user runs `cf upgrade-service`. A user can run `upgrade-service` even if an upgrade is not available. The broker should ultimately do a noop.

brokerapi/broker/decider/decider.go
FelisiaM: They result in different error messages so they should be kept separate to make sure there is a mre useful message

brokerapi/broker/decider/decider.go
```diff
@@ -19,61 +19,46 @@ const (
...
+		return Failed, apiresponses.ErrMaintenanceInfoConflict
+	case invalidUpdateAndMIChange, invalidUpdateAndMIRemoval:
+		// invalid mixing of an update with an attempt to change MI
+		return Failed, errInstanceMustBeUpgradedFirst()
+	case !requestHasUpdate && !details.MaintenanceInfoVersion.Equal(details.PreviousMaintenanceInfoVersion):
+		// MI changed and no updates, so must be an upgrade
 		return Upgrade, nil
+	default:
+		// It's not a valid or invalid upgrade, so must be an update
+		return Update, nil
```
FelisiaM, line 60: It looks like there is a usecase removed from the implementation and I'm not sure I could find the tests either. 
The one where the previous maintenance info is different from the catalog and no new MI has not been passed. 
This can happen when the broker and CAPI are out of sync (broker has been updated to have MI and has not been registered with CAPI yet). That should result on a `ErrMaintenanceInfoConflict` failure

FelisiaM (changes requested):
The validation of the previous MI has been removed in previous PR #545. I think that removed a check that is needed when the broker and CAPI get out of sync (which would be the case mainly for app users)

brokerapi/broker/decider/decider.go
```diff
@@ -19,61 +19,46 @@ const (
...
+	// There's an update to plan/parameters, new MI, and the new MI does not match the previous MI.
+	// So it's an attempt to upgrade and update at the same time.
+	invalidUpdateAndMIChange := requestHasUpdate && requestHasMI && !details.MaintenanceInfoVersion.Equal(details.PreviousMaintenanceInfoVersion)
+
+	// There's an update to plan/parameters, no new MI, plan does not have MI, but there's a previous MI value, so
+	// it looks like we are trying to change version (to having no MI) at the same time as an update.
+	// Valid options would be:
+	// - update to plan/parameters with previous MI the same as plan MI
+	// - upgrade to the no-MI plan with no updates of plan/parameters
+	invalidUpdateAndMIRemoval := requestHasUpdate && !requestHasMI && requestHasPreviousMI && planMaintenanceInfoVersion == nil
```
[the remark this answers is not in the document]
blgm: I've updated the comment to make it clearer. I did try out your suggestion, but this changed the error in other scenarios, which was an unwanted side-effect

brokerapi/broker/decider/decider.go
blgm: Good spot. I've added that back in.

brokerapi/broker/decider/decider.go
```diff
@@ -19,61 +19,56 @@ const (
...
+	case invalidUpdateAndMIChange, invalidUpdateAndMIRemoval:
+		// error: invalid mixing of an update with an attempt to change MI
+		return Failed, errInstanceMustBeUpgradedFirst()
+	case !requestHasMI && !requestHasUpdate && planMaintenanceInfoVersion == nil && details.PreviousMaintenanceInfoVersion != nil:
+		// MI removal: no MI in request because MI is being removed to match plan, and previous MI means it's a valid Upgrade
 		return Upgrade, nil
+	case requestHasMI && !requestHasUpdate && !details.MaintenanceInfoVersion.Equal(details.PreviousMaintenanceInfoVersion):
+		// add or change MI: MI changed and no updates, so must be an upgrade
+		return Upgrade, nil
+	case !requestHasMI && !planMaintenanceInfoVersion.Equal(details.PreviousMaintenanceInfoVersion):
```
FelisiaM, line 65: The request could include a plan change. In that case, the planMaintenanceInfoVersion would not be correct since it's referring to the new plan, not to the old one. I think the previous value needs to be compared to the MI of the previous plan

blgm: In the Cloud Service Broker, all plans are at the same maintenance info version. I could rename `planMaintenanceInfoVersion` to `serviceMaintenanceInfoVersion` to make that clearer. Would that help?

FelisiaM: I see your point. I think for now it's fine, and it would become a problem only if we ever decide to have different MI for different plans (let's hope we remember that)

FelisiaM (approved):

PR #551, opened 2022-06-23
feat: helper cf plugin - upgrade-all-service-instances

jimbo459:
### Checklist:


upgrade-all-plugin/internal/upgrader/upgrader.go
```diff
@@ -0,0 +1,136 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/cloudfoundry/cloud-service-broker/upgrade-all-plugin/internal/ccapi"
+	"github.com/cloudfoundry/cloud-service-broker/upgrade-all-plugin/internal/workers"
+)
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 . CCAPI
+type CCAPI interface {
```
FelisiaM, line 14: I found it confusing to have the interface and the implementation type called the same thing. Maybe it's worth renaming one of them e.g the interface can be `CFClient` (theoretically it can be fulfilled by a call to the API or to the CLI) so the interface should not care that it's calling an API?

upgrade-all-plugin/internal/upgrader/upgrader.go
```diff
@@ -0,0 +1,136 @@
...
+	GetServiceInstances([]string) ([]ccapi.ServiceInstance, error)
+	GetServicePlans(string) ([]ccapi.Plan, error)
+	UpgradeServiceInstance(string, string) error
+}
+
+func Upgrade(api CCAPI, brokerName string, batchSize int, l *log.Logger) error {
+	plans, err := api.GetServicePlans(brokerName)
```
FelisiaM, line 21: I think there is a chance here to exit early. If no plans were found for the broker we should exit. I don't think this would come as an error so we need one more check, but I might be worng

jimbo459: Yes you're right - this would actually fail on line 36 as we wouldn't be specifying any plan guids in the query we make there, but definitely better to fail earlier with a clean error message.

upgrade-all-plugin/internal/upgrader/upgrader.go
jimbo459: 👍 - makes sense!

upgrade-all-plugin/internal/upgrader/upgrader.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	l.Printf("Discovering service instances for broker: %s\n", brokerName)
+
+	serviceInstances, err := api.GetServiceInstances(planGUIDS)
+	if err != nil {
+		return err
+	}
+
+	type upgradeTask struct {
```
FelisiaM, line 41: nit: this is declared here, but not used until line 71. I would suggest to move it closer to where it gets used

upgrade-all-plugin/internal/upgrader/upgrader.go
```diff
@@ -0,0 +1,136 @@
...
+		GUID      string
```
FelisiaM, line 42: I know right now there is only one GUId, but I would prefer to have it named something like ServiceInstanceID to make it clear. Also MI to MaintenenaceInfo

upgrade-all-plugin/internal/ccapi/upgrade_instance.go
```diff
@@ -0,0 +1,42 @@
...
+
+			if si.LastOperation.State == "failed" && si.LastOperation.Type == "update" {
+				return fmt.Errorf("%s", si.LastOperation.Description)
+			}
+
+			if si.LastOperation.State != "in progress" || si.LastOperation.Type != "update" {
+				return nil
+			}
+		}
+		time.Sleep(time.Second)
```
FelisiaM, line 40: Are we risking inundating the API with too many requests. We know our upgrade process does take a while. Maybe its safer to ask every 10-30 sec. We know that CC polls the broker every 60sec by default, so we would not be having an updated result in a while

upgrade-all-plugin/internal/upgrader/upgrader.go
jimbo459: 👍 Yep good suggestion

upgrade-all-plugin/README.md
```diff
@@ -0,0 +1,25 @@
+## upgrade-all-service-instances (cf cli plugin)
+
+A CF-CLI plugin for upgrading all service instances in a CF foundation.
+
+### Purpose
+This tool was developed to allow operators to upgrade all service instances in a foundation, without having to navigate between orgs and spaces to discover upgradable instances.
```
FelisiaM, line 6: Just to remove confusion maybe we should say to allow users to upgrade all SIs they can access in a foundation. We can add that it's recommended to be used only by admins

upgrade-all-plugin/internal/ccapi/service_instances.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	var si serviceInstances
+	if err := c.requester.Get(fmt.Sprintf("v3/service_instances?per_page=5000&service_plan_guids=%s", strings.Join(planGUIDs, ",")), &si); err != nil {
+		return nil, fmt.Errorf("error getting service instances: %s", err)
+	}
+	return si.Instances, nil
+}
+
+type ServiceInstance struct {
```
FelisiaM, line 22: Nitpick: Just for consistency I think we should have the types at the top of the file

FelisiaM (approved):

PR #555, opened 2022-06-24
fix: brokerpaktestframework fails with "terraform_upgrade_path"

blgm:
The brokerpaktestframwork failed if a "terraform_upgrade_path" was
specified. This is because the framework used to update the manifest to
have a single version of Terraform at 1.1.4. But when other versions are
specified in the "terraform_upgrade_path" then the manifest validation
fails.

- The brokerpaktestframework now reads the Terraform versions present in
the brokerpak and fakes them all (with the default version)
- There is a test for the brokerpaktestframework which will hopefully
find some of the worst errors in future
- The signature of `FindServicePlanGUIDs()` has been changed to return
an error rather than fail silentlu
- The fake broker is now started on a random port so that it does not
interfere with any locally running CSB instances
- If building the fake brokerpak fails, an error is returned

[#181155154](https://www.pivotaltracker.com/story/show/181155154)

### Checklist:

### DIFF
--- a/brokerpaktestframework/broker_pack_path_finder.go
+++ b/brokerpaktestframework/broker_pack_path_finder.go
@@ -5,8 +5,16 @@ import (
 	"runtime"
 )
 
-func PathToBrokerPack() string {
-	_, file, _, _ := runtime.Caller(1)
+func PathToBrokerPack(skips ...int) string {
+	skip := 1
+	switch len(skips) {
+	case 0:
+	case 1:
+		skip += skips[0]
+	default:
+		panic("too many skips")
+	}
+	_, file, _, _ := runtime.Caller(skip)
 
 	return filepath.Dir(filepath.Dir(file))
 }
--- a/brokerpaktestframework/osbapi_helper.go
+++ b/brokerpaktestframework/osbapi_helper.go
@@ -1,6 +1,8 @@
 package brokerpaktestframework
 
 import (
+	"fmt"
+
 	"github.com/pivotal-cf/brokerapi/v8/domain"
 	"github.com/pivotal-cf/brokerapi/v8/domain/apiresponses"
 )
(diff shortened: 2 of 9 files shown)


brokerpaktestframework/broker_pack_path_finder.go
```diff
@@ -5,8 +5,16 @@ import (
...
+func PathToBrokerPack(skips ...int) string {
+	skip := 1
+	switch len(skips) {
+	case 0:
+	case 1:
+		skip += skips[0]
+	default:
+		panic("too many skips")
+	}
+	_, file, _, _ := runtime.Caller(skip)
```
zucchinidev, line 18: thoughts: we may want to make the comparison more evident:

```
	switch {
	case len(skips) == 1:
		skip += skips[0]
	case len(skips) > 1:
		panic("too many skips")
	}
```

brokerpaktestframework/osbapi_helper.go
```diff
@@ -27,12 +29,12 @@ func FindServicePlan(catalog *apiresponses.CatalogResponse, serviceName, service
...
+func FindServicePlanGUIDs(catalog *apiresponses.CatalogResponse, serviceName, planName string) (string, string, error) {
 	service := FindService(catalog, serviceName)
 	for _, plan := range service.Plans {
 		if plan.Name == planName {
-			return service.ID, plan.ID
+			return service.ID, plan.ID, nil
 		}
 	}
-	return "", ""
+	return "", "", fmt.Errorf("cannot find service %q and plan %q in catalog", serviceName, planName)
```
zucchinidev, line 39: praise: thank you very much George!

brokerpaktestframework/test_instance_builder.go
```diff
@@ -130,3 +125,12 @@ func writeManifest(brokerPackDir string, build string, workingDir string) (err e
...
 }
+
+func freePort() int {
+	listener, err := net.Listen("tcp", "localhost:0")
+	if err != nil {
+		panic("unable to open a listener port")
+	}
+	defer listener.Close()
+	return listener.Addr().(*net.TCPAddr).Port
+}
```
zucchinidev, line 136: praise: fantastic!!

integrationtest/brokerpaktestframework_test.go
```diff
@@ -0,0 +1,56 @@
+package integrationtest
```
zucchinidev, line 1: praise: thank you very much for this test!!

zucchinidev (approved):

jimbo459:
@blgm - did you mean to commit the `.gitkeep` file?

brokerpaktestframework/terraform_mock.go
```diff
@@ -7,12 +7,22 @@ import (
...
-
 	"github.com/onsi/gomega/gexec"
 )
 
-func NewTerraformMock() (TerraformMock, error) {
+func NewTerraformMock(opts ...Option) (TerraformMock, error) {
+	var version string
+
+	for _, o := range opts {
+		version = o()
```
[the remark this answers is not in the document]
blgm: Exactly. You could make this change later without breaking the API.

PR #581, opened 2022-07-18
fix: checks that terraform_upgrade_path terminates at default version

blgm:
A common error is to add a new default Terraform version and fail to
update the terraform_upgrade_path. This adds a check so that if the
terraform_upgrade_path does not terminate at the default version then
the brokerpak will fail to build, resulting in faster feedback.

### Checklist:


jimbo459:
This is a brilliant PR - thank you!

PR #584, opened 2022-07-27
feat: allow input fields to be nullable

blgm:
Fields have a type, for example number, string, boolean. Sometimes it's
useful for a field to also be `null`. This is allowed for HCL variables.
By introducing a `nullable` property, we allow fields to be set to
`null` in addition to values of the specified type.

[#182828491](https://www.pivotaltracker.com/story/show/182828491)

### Checklist:


docs/brokerpak-specification.md
```diff
@@ -323,17 +323,18 @@ The variable object describes a particular input or output variable. The
...
+| type*           | string            | The JSON type of the field. This MUST be a valid JSONSchema type excepting `null`.                                                                                                                                                                                                                                                                                                                    |
+| nullable*       | boolean           | Whether the field can be set to `null`. For example, if the type is `string`, this will allow the field to also be set to `null`, which is valid in HCL. This allows for differentiation between `null` and (say) an empty string.                                                                                                                                                                    |
```
jhvhs, line 331: *question:* What does the asterisk mean in this table?

blgm: That's a very good question. I've added a new commit that updates each table with a note underneath explaining what it is, and fixes the issue where `nullable` looks like it was required.

jhvhs (approved):

zucchinidev (approved):

PR #620, opened 2022-10-06
fix: set operation type in instance when updating

zucchinidev:
After executing a `deprovision` operation whose result was not satisfactory, the type of the last operation performed in the instance was stored in the database, that is, OperationType = deprovision, as established by the business logic. If after the `deprovision` operation, we performed an operation of updating a property of the same instance, for example modifying the property `deletion_protection` in some of the database services of some of the providers, the type of the last operation performed in the instance was not stored, that is, we did not assign the value `update` to the type of the last operation performed in the instance.

After a new call to endpoint /v2/service_instances/:instance_id/last_operation due to an asynchronous operation, the logic associated with this endpoint deleted the instance.

Steps we performed before fixing the problem:

* Deprovision operation executed successfully.
* OperationType = deprovision stored in our database associated with the instance.
* Update operation executed successfully in asynchronous mode.
* OperationType = update was not stored in our database. Instance with an incorrect state.
* LastOperation operation check executed successfully. Last OperationType = deprovision instead of update.
* Logic associated with deleting instance executed successfully.

[#183452044](https://www.pivotaltracker.com/story/show/183452044)

### Checklist:

* [x] Have you added Release Notes in the docs repositories?

### DIFF
--- a/brokerapi/broker/update_test.go
+++ b/brokerapi/broker/update_test.go
@@ -162,24 +162,6 @@ var _ = Describe("Update", func() {
 					OperationType: models.UpdateOperationType,
 					OperationID:   updateOperationID,
 				}, nil)
-
-				updateDetails = domain.UpdateDetails{
-					ServiceID: offeringID,
-					PlanID:    originalPlanID,
-					MaintenanceInfo: &domain.MaintenanceInfo{
-						Version: "2.0.0",
-					},
-					PreviousValues: domain.PreviousValues{
-						PlanID:    originalPlanID,
-						ServiceID: offeringID,
-						OrgID:     orgID,
-						SpaceID:   spaceID,
-						MaintenanceInfo: &domain.MaintenanceInfo{
-							Version: "2.0.0",
-						},
-					},
-					RawContext: json.RawMessage(fmt.Sprintf(`{"organization_guid":%q, "space_guid": %q}`, orgID, spaceID)),
-				}
 			})
 
 			It("should complete changing the instance operation type", func() {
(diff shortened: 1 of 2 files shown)


brokerapi/broker/update_test.go
```diff
@@ -199,8 +199,16 @@ var _ = Describe("Update", func() {
 				actualContext, _ := fakeServiceProvider.UpdateArgsForCall(0)
 				Expect(actualContext.Value(middlewares.OriginatingIdentityKey)).To(Equal(expectedHeader))
 
-				By("validating SI details is not updated")
-				Expect(fakeStorage.StoreServiceInstanceDetailsCallCount()).To(Equal(0))
+				By("validating SI details is updated")
+				storeServiceInstanceDetailsValidation(
```
pivotal-marcela-campo, line 203: Great check! I would just rename it to ExpectOperationTypeToBeUpdated or something alike

brokerapi/broker/update_test.go
```diff
@@ -199,8 +199,16 @@ var _ = Describe("Update", func() {
...
+					fakeStorage,
+					updateOperationID,
+					instanceID,
+					offeringID,
+					originalPlanID,
+					spaceID,
+					orgID,
+				)
 
 				By("validating provision parameters storing call")
```
pivotal-marcela-campo, line 213: Would remove this By, and probably just add the check to the previous method as it is all part of checking the operation was updated

brokerapi/broker/update_test.go
pivotal-marcela-campo: Same for the By comment - Validating SI operation type is updated would be more descriptive

brokerapi/broker/update_test.go
```diff
@@ -301,8 +311,16 @@ var _ = Describe("Update", func() {
 				Expect(actualVars.GetString("foo")).To(Equal("quz"))
 				Expect(actualVars.GetString("guz")).To(Equal("muz"))
 
-				By("validating SI details is not updated")
-				Expect(fakeStorage.StoreServiceInstanceDetailsCallCount()).To(Equal(0))
+				By("validating SI details is updated")
```
pivotal-marcela-campo, line 314: same comments as above

pivotal-marcela-campo (commented):
Great catch! just some naming/comment changes in the tests to make it more descriptive. The logic itself is perfect

pivotal-marcela-campo (approved):

PR #623, opened 2022-10-07
chore: remove superfluous IsSubsumePlan() method

blgm:
Contrary to its name, this method doesn't tell you whether a particular
plan is a subsume plan: it returns `true` if the plan has a subsume
field - whether or not that field is true or false. It might be more
accurate to call it IsPlanFromSubsumeServiceOffering(), but actually the
method is not needed. It's used as a guard around logic to extract
values from the output of `terraform show`. Values that should be read
are fields that have a `tf_attribute` value. Where there are no fields
with `tf_attribute` values, the `terraform show` will not be run.
There's therefore no need to have an additional guard from
IsSubsumePlan()

[#183463447](https://www.pivotaltracker.com/story/show/183463447)


pkg/providers/tf/imports.go
```diff
@@ -12,10 +12,6 @@ import (
 func (provider *TerraformProvider) GetImportedProperties(ctx context.Context, planGUID string, instanceGUID string, inputVariables []broker.BrokerVariable) (map[string]any, error) {
```
zucchinidev, line 12: thoughts: the parameter `planGUID` is no longer needed. We might want to create a separate PR with the next modifications:

- `pkg/broker/service_provider.go:45` `ServiceProvider` interface
- `brokerapi/broker/update.go:87` method call
- mock recreation
- pkg/providers/tf/imports_test.go

zucchinidev (approved):

PR #629, opened 2022-10-10
fix: make config parsing sane

nouseforaname:
if the broker uses a brokerpak that in turn uses
`${config("yaml.path.to.item")} HIL funtion, configs can only be parsed when the yaml value for `yaml.path.to.item` is a string. e.g.:

this will work
```
yaml.path.to.item: '{"key":"value"}'
```

this wont:

```
yaml.path.to.item: {"key":"value"}
```

this neither

```
yaml.path.to.item:
  key: value
```

we should support passing in valid yaml and json configuration via config files for items referenced via the hil func config.

### Checklist:

* ~[ ] Have you added Release Notes in the docs repositories?~

### DIFF
--- a/pkg/varcontext/interpolation/eval_test.go
+++ b/pkg/varcontext/interpolation/eval_test.go
@@ -33,38 +33,44 @@ func TestEval(t *testing.T) {
 		Expected      any
 		ErrorContains string
 	}{
-		"Non-Templated String":  {Template: "foo", Expected: "foo"},
-		"Basic Evaluation":      {Template: "${33}", Expected: "33"},
-		"Escaped Evaluation":    {Template: "$${33}", Expected: "${33}"},
-		"Missing Variable":      {Template: "${a}", ErrorContains: "unknown variable accessed: a"},
-		"Variable Substitution": {Template: "${foo}", Variables: map[string]any{"foo": 33}, Expected: "33"},
-		"Bad Template":          {Template: "${", ErrorContains: "expected expression"},
-		"Truncate Required":     {Template: `${str.truncate(2, "expression")}`, Expected: "ex"},
-		"Truncate Not Required": {Template: `${str.truncate(200, "expression")}`, Expected: "expression"},
-		"Counter":               {Template: "${counter.next()},${counter.next()},${counter.next()}", Expected: "1,2,3"},
-		"Query Escape":          {Template: `${str.queryEscape("hello world")}`, Expected: "hello+world"},
-		"Query Amp":             {Template: `${str.queryEscape("hello&world")}`, Expected: "hello%26world"},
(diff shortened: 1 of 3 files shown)


pkg/varcontext/interpolation/funcs.go
```diff
@@ -59,8 +59,18 @@ func hilFuncConfig() ast.Function {
...
+			key := args[0].(string)
+			if viper.IsSet(key) {
+				val := viper.Get(key)
+				// Check If we're handling a nested object
+				if mapVal, ok := val.(map[string]interface{}); ok {
+					bytes, err := json.Marshal(mapVal)
+					// return string representation of the object contents
+					return string(bytes), err
+				}
+				// return string and remove leading/trailing whitespace
```
blgm, line 72: The comment says "remove leading/trailing whitespace" but the code does "remove leading/trailing newline". What's the intention? If the intention is to remove whitespace, then the TrimSpace() function can be helpful.

pkg/varcontext/interpolation/eval_test.go
```diff
@@ -33,45 +33,54 @@ func TestEval(t *testing.T) {
...
+		"env var":                            {Template: `${env("FOO")}`, Expected: `Bar`},
+		"missing env var":                    {Template: `${env("_MISSING")}`, ErrorContains: "missing environment variable _MISSING"},
+		"config val":                         {Template: `${config("config.val")}`, Expected: `foo`},
+		"config val nested map":              {Template: `${config("test.map")}`, Expected: `{"value":"one"}`},
+		"config val nested map nested value": {Template: `${config("test.map.value")}`, Expected: `one`},
+		"config val nested json object":      {Template: `${config("test.object")}`, Expected: `{"value":"one"}`},
+		"config val nested json object nested value": {Template: `${config("test.object.value")}`, Expected: `one`},
+		"config val nested string object":            {Template: `${config("test.string_object")}`, Expected: `{"value":"one"}`},
+		"config val nested multiline string object":  {Template: `${config("test.string_multiline_object")}`, Expected: `{"value":"one"}`},
+		//		"config val nested object": {Template: `${config("test.object.value")}`, Expected: `one`},
```
blgm, line 73: Is the commented out testcase a duplicate that can be removed?

blgm (changes requested):
This is a worthwhile improvement. There are a couple of comments that need to be tidied up.

pkg/varcontext/interpolation/funcs.go
nouseforaname: nice catch, TrimSpace was intended

pkg/varcontext/interpolation/eval_test.go
nouseforaname: yep, copy paste dupe. Thanks

blgm (approved):

PR #631, opened 2022-10-11
fix: ability to skip reading tf_attributes for optional resources

blgm:
The `tf_attribute` feature is used to "read back" values from Terraform in cases where resources have been subsumed, and property values may not have been stored. This ensures that resources are not overwritten during subsequent updates. Sometimes the existence of a resource is optional. For example the Azure brokerpak "csb-azure-mssql-db-failover-group" offering has an "existing" flag that will use an existing failover group rather than create one using Terraform. In this case, the reading of "tf_attribute" needs to be skipped because it will otherwise fail as there is nothing to read. The new "tf_attribute_skip" field allows reading of the "tf_attribute" to be optionally skipped depending on the value of another field.

[#183463447](https://www.pivotaltracker.com/story/show/183463447)


brokerapi/broker/update.go
```diff
@@ -84,12 +84,16 @@ func (broker *ServiceBroker) Update(ctx context.Context, instanceID string, deta
...
-		return domain.UpdateServiceSpec{}, fmt.Errorf("error retrieving subsume parameters for %q: %w", instanceID, err)
+		return domain.UpdateServiceSpec{}, err
+	}
+	importedParams, err := serviceProvider.GetImportedProperties(ctx, instance.GUID, serviceDefinition.ProvisionInputVariables, initialProperties)
+	if err != nil {
+		return domain.UpdateServiceSpec{}, fmt.Errorf("error retrieving expected parameters for %q: %w", instanceID, err)
 	}
 
-	mergedDetails, err := mergeJSON(provisionDetails, parsedDetails.RequestParams, importedParams)
+	mergedDetails, err := mergeJSON(provisionDetails, importedParams, parsedDetails.RequestParams)
```
zucchinidev, line 96: question: @blgm, wouldn't be enough passing `importedParams` and `initialProperties`?

```
mergedDetails, err := mergeJSON(initialProperties, importedParams)
```

integrationtest/tf_attribute_skip_test.go
```diff
@@ -0,0 +1,64 @@
...
+		Expect(receiver.Description).To(Equal(fmt.Sprintf(`error retrieving expected parameters for "%s": cannot find required import values for fields: does.not.exist`, s.GUID)))
+	})
+
+	It("can skip based on a stored request parameter", func() {
+		s := testHelper.Provision(serviceOfferingGUID, defaultPlanGUID, `{"skip":true}`)
+		testHelper.UpdateService(s)
+	})
+
+	It("can skip based on a new request parameter", func() {
+		s := testHelper.Provision(serviceOfferingGUID, defaultPlanGUID, nil)
```
zucchinidev, line 55: thoughts: I guess passing `nil` or not passing anything makes no difference

integrationtest/tf_attribute_skip_test.go
```diff
@@ -0,0 +1,64 @@
...
+		testHelper.UpdateService(s, `{"skip":true}`)
+	})
+
+	It("can skip based on a plan parameter", func() {
+		const skipPlanGUID = "56591d42-48af-11ed-bda0-0327763028ca"
+		s := testHelper.Provision(serviceOfferingGUID, skipPlanGUID, nil)
```
zucchinidev, line 61: same comment as the previous

pkg/broker/service_provider.go
```diff
@@ -42,7 +42,7 @@ type ServiceProvider interface {
 	UpgradeBindings(ctx context.Context, instanceContext *varcontext.VarContext, bindingContexts []*varcontext.VarContext) error
 
 	// GetImportedProperties extracts properties that should have been saved as part of subsume operation
-	GetImportedProperties(ctx context.Context, planGUID string, instanceGUID string, inputVariables []BrokerVariable) (map[string]any, error)
+	GetImportedProperties(ctx context.Context, instanceGUID string, inputVariables []BrokerVariable, initialProperties map[string]any) (map[string]any, error)
```
zucchinidev, line 45: praise: thanks for deleting the `planGUID` parameter

pkg/providers/tf/imports.go
```diff
@@ -30,15 +33,31 @@ func (provider *TerraformProvider) GetImportedProperties(ctx context.Context, pl
...
+			// whole varcontext, so some fields may be missing their default
+			// values. When the attribute can't be looked up, we act as if
+			// it was set to "false". This is because we don't want an error
+			// when the value of the field would simply default to false if
+			// the whole varcontext had been computed. This does mean that
+			// a reference to an invalid field won't fail, but that's better
+			// than introducing unexpected errors.
+			if skip, _ := initialProperties[vars.TFAttributeSkip].(bool); skip {
+				continue
+			}
```
zucchinidev, line 54: thoughts: it may be more readable using:

```
if skip, ok := initialProperties[vars.TFAttributeSkip].(bool); !ok || skip {
          continue
}
```

zucchinidev (approved):
I’ve asked a question, @blgm , but it’s not blocking.

brokerapi/broker/update.go
blgm: I think not because `initialProperties` also has values from `plan.GetServiceProperties()` which are constants specific to the plan, and we don't want to save them as parameters that the user has specifically requested.

zucchinidev: Thank you very much for your answer!

PR #649, opened 2022-10-29
chore: common helper for brokerpaktestframework & integrationtest

blgm:
brokerpaktestframework and integrationtest perform similar actions of
building a brokerpak, starting a broker and performing OSBAPI actions.
Up to now they have had separate but similar helpers. Now we have
introduced the "testdrive" helper which can work for both of them,
allowing a lot of near-duplicate code to be removed.

Note that this is a technically breaking change because it makes some small changes to the API of brokerpaktestframework - specifically removing the `BrokerURL`/`BrokerUrl` methods. But these are not actually called in the AWS, Azure or GCP brokerpaks, so it doesn't actually break any of the consumers.


blgm:
Superseded by #661  and #662

PR #660, opened 2022-11-08
feat: local development experience

blgm:
**NOTE:** #652 should be merged first

Building on the "csb create-service" experience, this feature enables
the full lifecycle of mimic CF CLI development experience.

The aim is to have commands that look familiar, and allow fast
prototyping and testing of brokerpaks by skipping the need to push the
CSB App to CloudFoundry.

Commands enabled are:

- csb create-service (as before, but can now take parameters)
- csb services (to list created services)
- csb update-service (will allow the no-op update)
- csb delete-service (to clean up)
- csb create-service-key (creates a "binding")
- csb delete-service-key (deletes a "binding")
- csb service-keys (lists "bindings")
- csb marketplace (lists available services and plans)


cmd/create_service.go
```diff
@@ -1,25 +1,100 @@
...
+	"github.com/cloudfoundry/cloud-service-broker/internal/local"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-
-	"github.com/cloudfoundry/cloud-service-broker/internal/createservice"
 )
 
 func init() {
-	var c string
+	var (
```
FelisiaM, line 10: name of the file no longer matches the commands list

FelisiaM: will merge all prs then can update the name

PR #680, opened 2022-11-25
chore: factor out brokerpack packer from testdrive package

blgm:
The 'testdrive' package is shared between integration tests and the CSB local experience. But the brokerpak builder was used only by the integration tests. So this part of the package has been moved to become a test helper for the integration tests, keeping the testdrive package more focused.


integrationtest/info_endpoint_test.go
```diff
@@ -16,9 +17,9 @@ var _ = Describe("Info Endpoint", func() {
 	var broker *testdrive.Broker
 
 	BeforeEach(func() {
-		brokerpak, err := testdrive.BuildBrokerpak(csb, fixtures("info-endpoint"))
-		Expect(err).NotTo(HaveOccurred())
+		brokerpak := must(packer.BuildBrokerpak(csb, fixtures("info-endpoint")))
 
+		var err error
```
jhvhs, line 22: **nitpick**: extra var declaration can be avoided via use of `must` helper.

jhvhs (changes requested):

integrationtest/info_endpoint_test.go
blgm: Absolutely. That's updated.

jhvhs (approved):

PR #694, opened 2023-01-04
fix: return HTTP 410 when deleting non-existent binding

blgm:
The Open Service Broker API (OSBAPI) spec says that an HTTP 410 should be returned when trying to delete a binding that does not exist.

[#183405838](https://www.pivotaltracker.com/story/show/183405838)

### DIFF
--- a/brokerapi/broker/bind_test.go
+++ b/brokerapi/broker/bind_test.go
@@ -228,7 +228,7 @@ var _ = Describe("Bind", func() {
 				fakeStorage.ExistsServiceBindingCredentialsReturns(true, nil)
 			})
 
-			It("should error", func() {
+			It("should return HTTP 409 as per OSBAPI spec", func() {
 				_, err := serviceBroker.Bind(context.TODO(), instanceID, bindingID, bindDetails, false)
 
 				Expect(err).To(MatchError(apiresponses.ErrBindingAlreadyExists))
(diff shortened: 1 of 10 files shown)


brokerapi/broker/bind_test.go
```diff
@@ -228,7 +228,7 @@ var _ = Describe("Bind", func() {
 				fakeStorage.ExistsServiceBindingCredentialsReturns(true, nil)
 			})
 
-			It("should error", func() {
+			It("should return HTTP 409 as per OSBAPI spec", func() {
```
zucchinidev, line 231: praise: thank you very much!

brokerapi/broker/unbind.go
```diff
@@ -50,7 +40,18 @@ func (broker *ServiceBroker) Unbind(ctx context.Context, instanceID, bindingID s
 	// get existing service instance details
 	instance, err := broker.store.GetServiceInstanceDetails(instanceID)
 	if err != nil {
-		return domain.UnbindSpec{}, fmt.Errorf("error retrieving service instance details: %s", err)
+		// in practice, we will not see this error as the binding will likely not exist
+		return domain.UnbindSpec{}, apiresponses.ErrInstanceDoesNotExist
```
zucchinidev, line 44: thoughts: @blgm, I see everything right except this little piece. I’m not saying it’s wrong, but I do see possible side effects. I understand the intent, but I think we’d be hiding networking errors, or errors decoding the JSON. I approve the PR and let your choice change this part or not

blgm: I can see arguments on both sides. I decided to stick with what I did because I have a slight preference for delete operations that don't fail, as failing delete operations leave you in a situation that's hard to clean up - since the action you take to clean up just fails.

zucchinidev: Thanks @blgm!!

PR #696, opened 2023-01-05
fix: csb local should filter out unspecified environment variables

blgm:
We got some feedback when someone was running the csb local command that there was a failure because the user had CH_CRED_HUB_URL set, and the CSB local experience does not have CredHub available.

The command has now been improved so that it now only allows certain specified environment variables, which are the ones from the brokerpak manifest and any than start "GSB_"


internal/testdrive/broker_start.go
```diff
@@ -104,7 +101,25 @@ func StartBroker(csbPath, bpk, db string, opts ...StartBrokerOption) (*Broker, e
...
+	a := make(map[string]struct{})
+	for _, allow := range allowed {
+		a[allow] = struct{}{}
+	}
+
+	return func(cfg *startBrokerConfig) {
+		var result []string
+		for _, e := range cfg.env {
+			name := varname(e)
+			if _, ok := a[name]; ok || strings.HasPrefix(name, "GSB_") {
```
pivotal-marcela-campo, line 118: It looks like some broker variables wouldn't be configured, like encryption ones. I think it is okay anyway for local development and we can always add them if desperate to test any of that locally. Just wanted to point it out.

pivotal-marcela-campo (approved):
Amazing!

internal/testdrive/broker_start.go
blgm: Yes, this is one of the downsides. I was keen to go for an allow-list of variables as it feels like a more controlled environment than just having a deny-list for the CH_* variables. But the downside is that the list requires more maintenance. There's definitely more work to do in this area.

PR #714, opened 2023-01-25
feat: example tests automatically start broker

blgm:
Previously, to run the example tests, one had to start a broker in one terminal, and run the "csb client run-examples" command in another terminal. This feature builds on the CSB local experience and automatically starts the broker for you. The new commands are:

- csb examples (to list examples)
- csb run-examples (to run examples)

This feature is experimental for now, and the old functionality still exists. Assuming the new way works well, we should eventually remove the old way.

### DIFF
--- a/internal/local/run_examples.go
+++ b/internal/local/run_examples.go
@@ -1,0 +1,23 @@
+package local
+
+import (
+	"log"
+
+	"github.com/cloudfoundry/cloud-service-broker/pkg/client"
+)
+
+func RunExamples(serviceOfferingName, exampleName, cachePath string) {
+	pakDir, cleanup := pack(cachePath)
+	defer cleanup()
+
+	broker := startBroker(pakDir)
+	defer broker.Stop()
+
+	examples, err := listExamples(broker.Port)
+	if err != nil {
+		log.Fatal(err)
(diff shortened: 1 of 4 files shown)


internal/local/run_examples.go
```diff
@@ -0,0 +1,23 @@
...
+
+	broker := startBroker(pakDir)
+	defer broker.Stop()
+
+	examples, err := listExamples(broker.Port)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	const jobCount = 1_000_000
```
zucchinidev, line 21: question: I wouldn't create a million of workers. I rather do `g := runtime.NumCPU()` instead of using a magic number or I would do a load testing and try to adjust the number. Do you have any reason to select that number?

blgm: If the tests were processor-intensive then I would agree. But they actually spend most of their time waiting on the IaaS. And the IaaS can handle a lot of parallelisation. Take the example of the Azure brokerpak, it has 35 example tests, and in our CI we run them all in parallel. Maybe we will have to tune this number down eventually, but my idea was to set it high to start with, otherwise the tests just take longer than they need to.

PR #721, opened 2023-02-01
DO NOT MERGE feat: ability to stop merge of bind outputs with provision outputs

blgm:
**DON'T MERGE YET, WE NEED A TEAM DISCUSSION ABOUT THIS**
    
There is a feature flag `CSB_DISABLE_BIND_OUTPUT_MERGING` that will
stop the outputs from the provision operation from appearing in binding
credentials.
    
Up until now, binding credentials have been a merge of Terraform outputs
from the bind operation and Terraform outputs from the provision
operation.
    
This is behaviour is undesirable because:
- many brokerpaks explicitly copy data from the instance to the credentials, and in general it's better to have one way of doing things
- explicitly copying data is easier to reason about, many folks found this behaviour unexpected (although it was documented)
- it's easy to make a security mistake. At the moment many brokerpaks create an admin user during a provision operation, and mask the values in the bind (e.g. by overwriting username/password with a non-admin user). An innocent looking rename (e.g. password -> admin_password) could have the effect of leaking admin credentials in a binding.
    
In the future, this may be made the default behaviour.

[#184269641](https://www.pivotaltracker.com/story/show/184269641)

### DIFF
--- a/brokerapi/broker/bind.go
+++ b/brokerapi/broker/bind.go
@@ -151,10 +151,11 @@ func validateBindParameters(params map[string]any, validUserInputFields []broker
 // BuildInstanceCredentials combines the bind credentials with the connection
 // information in the instance details to get a full set of connection details.
 func buildInstanceCredentials(credentials map[string]any, outputs storage.JSONObject) (*domain.Binding, error) {
-	vc, err := varcontext.Builder().
-		MergeMap(outputs).
-		MergeMap(credentials).
-		Build()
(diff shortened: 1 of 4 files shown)


blgm (commented):
On reflection, this approach involves a lot of copy/paste boilerplate where bind outputs are just duplicates of provision outputs. It was suggested that adding a property to provision outputs like `expose_in_binding` or similar could fix the issues that we are having in a more concise way.

If we did this, we would likely want `expose_in_binding` to be false by default, in which case it should be controlled by a feature flag.

PR #722, opened 2023-02-02
chore: remove usage of deprecated constants

blgm:
The very latest staticcheck picks up on some more deprecation warnings, and this change fixes the warnings.


pivotal-marcela-campo (approved):
Thanks!

PR #736, opened 2023-02-14
fix: broker should not start when no services are definied

blgm:
From time to time, someone misconfigures the broker so that no services are defined. When this happens, the broker starts successfully but when registering to CloudFoundry the "cf create-service-broker" command fails. Often folks spend a lot of time investigating CloudFoundry, when an earlier failure would have directed them to the source of the problem. This commit introduces the earlier failure by refusing to start when no services are defined.

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -107,10 +107,14 @@ func serve() {
 	}
 
 	services, err := serviceBroker.Services(context.Background())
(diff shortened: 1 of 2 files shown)


zucchinidev (approved):

zucchinidev:
Thank you very much @blgm

PR #758, opened 2023-02-27
fix: SQLite3 concurrency issues when running examples

blgm:
SQLite3 does not support performing similar operations (e.g. INSERT) concurrently on the same table from the same connection. As CSB uses the same database connection for all goroutines, we were occasionally seeing errors saying "database is locked".

Enabling the full mutex mode will serialise database operations and hopefully stop this error from occurring.


pivotal-marcela-campo (approved):
Interesting, thanks!

PR #778, opened 2023-03-15
fix: invalid cache state causing invalid brokerpaks

blgm:
When using a cache to build brokerpaks, a source path is converted into an MD5 sum and this is used to name a directory in the cache, which is populated by the contents. It was observed that cache directories were sometimes empty - presumably because the operating system was cleaning up files in /tmp. But because the cache directory still existed, it was still considered valid, and a brokerpak would be successfully built, but with missing content. This commit introduces an additional check that the cache directory should not be empty, which should resolve this kind of error.


pivotal-marcela-campo (approved):
Thanks!

PR #785, opened 2023-03-17
fix: unbind should succeed CredHub cleanup fails

blgm:
When CredHub is enabled, the CSB will store binding credentials in CredHub. During the unbind operation, it will try to remove those credentials. In most cases, the credentials will have been invalidated first, but this depends on the service implementation.

The cleanup can fail for a number of reasons:
- the service name has changed
- CredHub has already been cleaned up
- there's a failure in CredHub

In general we prefer deletion operations (such as unbind) to be idempotent so that they can be retried. We don't want to have to advise people to add a credential back into CredHub in order to make an unbind operation succeed. So if there is a failure to delete a credential from CredHub, we will succeed and log the failure. This mirrors logic when we can't remove the permissions for the credential. This is a tradeoff, and it would be nicer if CredHub were also idempotent.

[#184672112](https://www.pivotaltracker.com/story/show/184672112)


brokerapi/broker/unbind.go
```diff
@@ -76,17 +76,14 @@ func (broker *ServiceBroker) Unbind(ctx context.Context, instanceID, bindingID s
...
-			broker.Logger.Error(fmt.Sprintf("fail to delete permissions on the key %s", credentialName), err)
+		if err := broker.Credstore.DeletePermission(credentialName); err != nil {
+			broker.Logger.Error(fmt.Sprintf("failed to delete permissions on the CredHub key %s", credentialName), err)
 		}
 
-		err := broker.Credstore.Delete(credentialName)
-		if err != nil {
-			return domain.UnbindSpec{}, err
+		if err := broker.Credstore.Delete(credentialName); err != nil {
+			broker.Logger.Error(fmt.Sprintf("failed to delete CredHub key %s", credentialName), err)
```
pivotal-marcela-campo, line 86: Looks good, my only question is what happens if the error is not an NotFound but a connection error. We would be leaving orphaned credentials in that scenario as well. Might be good enhancing this handling

pivotal-marcela-campo (approved):

brokerapi/broker/unbind.go
blgm: This is a fair comment. I'm not sure we have a good enough API to be able to differentiate in a sustainable way. But it's something to consider improving the in the future.

PR #892, opened 2023-10-31
chore: physically delete records from the database on

pivotal-marcela-campo:
deprovision/unbind

Records where being soft-deleted or not even soft-deleted from the database on deprovision and unbind.
The terraform_deployments table (which was not even being soft-deleted) would accumulate thousands of records over time, making consistency checks very slow and causing cloud foundry to time out on app start up. Although we could work on changing the way the checks are performed, the broker doesn't have any uses for these soft-deleted or old records once the operation has completed successfully. Hence there is no real use at the moment in keeping this data in the database and take over space and affect performance.

[#186230750](https://www.pivotaltracker.com/story/show/186230750)

### Checklist:

### DIFF
--- a/internal/storage/provision_request_details_test.go
+++ b/internal/storage/provision_request_details_test.go
@@ -108,7 +108,7 @@ var _ = Describe("ProvisionRequestDetails", func() {
 			addFakeProvisionRequestDetails()
 		})
 
-		It("deletes psychically from the database", func() {
+		It("deletes physically from the database", func() {
 			exists := func() bool {
 				var count int64
 				Expect(db.Model(&models.ProvisionRequestDetails{}).Unscoped().Where(`service_instance_id="fake-instance-id"`).Count(&count).Error).NotTo(HaveOccurred())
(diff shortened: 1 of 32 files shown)


internal/storage/provision_request_details_test.go
```diff
@@ -108,10 +108,10 @@ var _ = Describe("ProvisionRequestDetails", func() {
 			addFakeProvisionRequestDetails()
 		})
 
-		It("deletes from the database", func() {
+		It("deletes psychically from the database", func() {
```
blgm, line 111: Impressive programming. Or perhaps a typo?

blgm (approved):
Looks good. An improvement for the future would be to use actual relationships with the data so that we can make use of the database doing a cascade delete. It feels like we are writing code here that's already implemented in the database, and I think that's the right thing to do in the short term, but in the long term we can refactor and reduce the amount of code that we need to maintain.

internal/storage/provision_request_details_test.go
pivotal-marcela-campo, line 111:
```suggestion
		It("deletes physically from the database", func() {
```

PR #894, opened 2023-11-07
feat: `pak init` to include usable GUIDs in the template

jhvhs:
Previously, the newly created brokerpak had `00000000-0000-0000-0000-000000000000` as the service GUID, and
`00000000-0000-0000-0000-000000000001` as the plan GUID. This would require the author to re-generate the above values. The probability of GUIDs clashing in this case is negligible.

### Checklist:

* [~] Have you added or updated tests to validate the changed functionality?
* [~] Have you added Release Notes in the docs repositories?


pivotal-marcela-campo (approved):
Thanks!!

PR #899, opened 2023-11-09
chore(csb): csb update command supports upgrades enabled

pivotal-marcela-campo:
### Checklist:

~~* [ ] Have you added Release Notes in the docs repositories?~~

### DIFF
--- a/internal/local/update_service.go
+++ b/internal/local/update_service.go
@@ -1,9 +1,10 @@
 package local
 
 import (
-	"log"
-
+	"fmt"
 	"github.com/cloudfoundry/cloud-service-broker/internal/testdrive"
+	"github.com/pivotal-cf/brokerapi/v10/domain"
+	"log"
 )
 
 func UpdateService(name, plan, params, cachePath string) {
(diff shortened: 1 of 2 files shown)


internal/local/update_service.go
```diff
@@ -1,21 +1,27 @@
...
 )
 
 func UpdateService(name, plan, params, cachePath string) {
 	pakDir, cleanup := pack(cachePath)
 	defer cleanup()
 
 	serviceInstance := lookupServiceInstanceByGUID(nameToID(name))
-
+	deployment, _ := store().GetTerraformDeployment(fmt.Sprintf("tf:%s:", serviceInstance.GUID))
+	tfVersion, _ := deployment.TFWorkspace().StateTFVersion()
```
zucchinidev, line 18:
```suggestion
	deployment, err := store().GetTerraformDeployment(fmt.Sprintf("tf:%s:", serviceInstance.GUID))
	if err != nil {
		log.Fatal(err)
	}
	tfVersion, err := deployment.TFWorkspace().StateTFVersion()
	if err != nil {
		log.Fatal(err)
	}
```

zucchinidev: I would prefer to log the error, if you think that by the characteristics of the execution it is not possible to fail, then feel free to ignore the suggestion

PR #955, opened 2024-02-15
fix: successfully delete instances and bindings with invalid state

blgm:
NOTE: please merge PR #954 first

As part of the upgrade epic, we added checks to make sure that service
instances and bindings had been updated to the latest version. This
involves parsing the Terraform state and extracting the version. But
sometimes the Terraform state is not valid due to being empty, or being
invalid JSON. Typically this is after a failure to provision/bind,
and the service instance or binding is in a failed state. In these
situations, the lack of state data means CSB does not have enough data to
successfully clean up using Terraform. We have had feedback that this
can be problematic for users because they have to do a manual purge of
the service instance, and it would be easier if the delete operations
would just work. This is a tradeoff because a manual purge could have
the effect of triggering someone to look at the backing IaaS to look for
resource leakage. By making this just work, we might make resource
leakage worse.

[#185835561](https://www.pivotaltracker.com/story/show/185835561)

### DIFF
--- a/brokerapi/broker/deprovision.go
+++ b/brokerapi/broker/deprovision.go
@@ -2,16 +2,18 @@ package broker
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"code.cloudfoundry.org/lager/v3"
-	"github.com/pivotal-cf/brokerapi/v10/domain"
-	"github.com/pivotal-cf/brokerapi/v10/domain/apiresponses"
-
 	"github.com/cloudfoundry/cloud-service-broker/dbservice/models"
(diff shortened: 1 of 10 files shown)


pivotal-marcela-campo (approved):

pivotal-marcela-campo:
LGTM. There is one scenario missing however, as `serviceProvider.CheckOperationConstraints` would fail before if there is no deployment at all, and this [indeed can happen ](https://github.com/cloudfoundry/cloud-service-broker/blob/cd9f3cf87cb6eb88f21c4ae3dc398237d6520276/pkg/providers/tf/provider.go#L79). Will PR that check separately.

PR #968, opened 2024-03-12
chore: improve help command output

pivotal-marcela-campo:
### Checklist:

~~* [ ] Have you added Release Notes in the docs repositories?~~

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -23,8 +23,9 @@ import (
 
 func init() {
 	configCmd := &cobra.Command{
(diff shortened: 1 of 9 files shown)


blgm (approved):
Happy for this to be merged as it is. If I were being very fussy I would have suggested that `"local"` were turned into a constant rather than a repeated string, and rather than having a var that's used only once, I'd prefer to inline it like:
```go
rootCmd.AddGroup(&cobra.Group{...})
```

PR #980, opened 2024-03-26
test: recovery after a broker restart

blgm:
We know that recovering in-progress operations after a broker restart is a non-ideal experience. These tests are to capture the current state of the world, so that we can see an improvement when we fix the problems.

[#178142867](https://www.pivotaltracker.com/story/show/178142867)


integrationtest/termination_recovery_test.go
```diff
@@ -0,0 +1,160 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		Expect(lastOperation.Description).To(Equal("provision in progress"))
+		Expect(lastOperation.State).To(BeEquivalentTo("in progress"))
+
+		// OSBAPI requires that HTTP 409 (Conflict) is returned
+		By("refusing to allow a duplicate instance")
+		response = broker.Client.Provision(instanceGUID, serviceOfferingGUID, servicePlanGUID, uuid.New(), nil)
+		Expect(response.Error).NotTo(HaveOccurred())
+		Expect(response.StatusCode).To(Equal(http.StatusConflict))
+
```
pivotal-marcela-campo, line 64: Out of curiosity, what happens here if we try to update?

pivotal-marcela-campo (approved):

integrationtest/termination_recovery_test.go
blgm: It fails with an internal server error and:
```
{"timestamp":"1711540247.853883982","source":"cloud-service-broker","message":"cloud-service-broker.unknown-error","log_level":2,"data":{"error":"terraform version check failed: workspace state not generated"}}
```
In a delete we have a special case that ignores this error allowing the cleanup. But we don't do that for update, and failing is probably the right thing to do here.

PR #983, opened 2024-03-27
fix: recover in-progress operations on broker restart

blgm:
- previously if an operation was in progress when the broker terminated it would forever be recorded as "in progress" in the broker database and was hard to clean up
- now on startup, the broker marks any in-progress operations as failed
- this allows the service instance to be cleaned up without intervention from a CloudFoundry administrator

[#178142867](https://www.pivotaltracker.com/story/show/178142867)

### Checklist:

### DIFF
--- a/dbservice/recover_in_progress_operations_test.go
+++ b/dbservice/recover_in_progress_operations_test.go
@@ -1,0 +1,83 @@
+package dbservice
+
+import (
+	"strings"
+	"testing"
+
+	"code.cloudfoundry.org/lager/v3/lagertest"
+	"github.com/cloudfoundry/cloud-service-broker/dbservice/models"
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+)
+
+func TestRecoverInProgressOperations(t *testing.T) {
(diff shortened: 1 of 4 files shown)


dbservice/recover_in_progress_operations_test.go
```diff
@@ -0,0 +1,83 @@
+package dbservice
```
pivotal-marcela-campo, line 1: non-blocking: we have been adding ginkgo tests in other areas and transforming go tests to ginkgo when we could. Adding a new test with go test makes it more difficult to progress to having all the tests in the same framework. non-blocking as I understand there could be some advantages in having all the tests in a package written in the same way and it might be a matter of taste. I would prefer new tests are written in ginkgo regardless.

pivotal-marcela-campo (approved):

dbservice/recover_in_progress_operations_test.go
blgm: Addressed in subseqent PR: https://github.com/cloudfoundry/cloud-service-broker/pull/1000

PR #1001, opened 2024-04-08
fix: change tofu unpacking to default to terraform registry

pivotal-marcela-campo:
We are keeping the default provider registry as terraform's own registry for now instead
of switching to OpenTofu's. We had done this during `pak` but had still defaulting to opentofu registry
when unpacking which caused 
* the state to contained references to opentofu's registry instead, and 
* the terraform provider to unpak under a `registry.opentofu.org` directory instead of the expected `registry.terraform.io` directory.

With this change we expect to have only references to registry.terraform.io unless otherwise specified by the brokerpak both in 
* the state and 
* directory structure

### Checklist:


go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/hashicorp/hil v0.0.0-20210521165536-27a72121fd40
 	github.com/heptiolabs/healthcheck v0.0.0-20180807145615-6ff867650f40
 	github.com/maxbrunsfeld/counterfeiter/v6 v6.8.1
+	github.com/onsi/ginkgo v1.16.5
```
blgm, line 17: Looks like Ginkgo V1 has been re-introduced

blgm (changes requested):
In the comment:
```
We are keeping the default provider registry as terraform's own registry for now instead
of switching to terraform.
```
Presumably the last word should be OpenTofu?

go.mod
pivotal-marcela-campo: fixed

FelisiaM (approved):

blgm (approved):

PR #1012, opened 2024-04-25
fix: only run database startup when running server

blgm:
An error was discovered where database startup, including database migrations and failed operation recovery was being run for every command that accessed the database, including "tf dump". This had the effect of marking any in-progress operations as failed, even if they were still really in progress. This has been a problem since the introduction of in-progress operation recovery in v1.1.0 (commit c8452677ef3d141efd150c5adbd4392377e2439c).

[#187500102](https://www.pivotaltracker.com/story/show/187500102)


pivotal-marcela-campo (approved):
We did miss that! Thanks for fixing!

PR #1014, opened 2024-04-26
feat!: remove subsume/import function

blgm:
The idea behind subsume was that it enabled pre-existing services to be taken under management by the CSB. In reality it proved quite unreliable as the pre-existing services had to very closely match the definition in the brokerpak. As a result it was not widely used and we are not aware of any brokerpaks that use subsume any more. So we have taken the decision to remove the function from CSB as it allows the code base to be simplified.


zucchinidev (approved):
Great job, @blgm 

I can't spot anything in the PR that makes me think we're introducing a bug, but there are many changes, so I hope I reviewed the files thoroughly.

I want to mention something as a reflection: we are modifying the workspace structure, and we store the data representing that structure in our database. This means that if we develop something that reads and represents old data, we will likely encounter errors when extracting the data. I don't believe we're doing anything similar right now, and I hope we won't, but I wanted to note this as a future reminder.

- There is a reference to `tf_attribute` in the documentation that we need to delete: `docs/brokerpak-specification.md:332`

blgm:
Thanks @zucchinidev. The risks around change in the data structures is noted. I think the JSON parser will just ignore them in default mode. I did an upgrade test and didn't see any issues, but I agree that it's one of the riskier parts of this change.

I've fixed the additional `tf_attribute` in the docs.

PR #1025, opened 2024-05-24
chore: debug option to leave workspace on disk

blgm:


pkg/providers/tf/workspace/workspace.go
```diff
@@ -25,18 +25,27 @@ import (
...
+	binaryName                       = "tofu"
+	leaveWorkspaceDirectoryConfigKey = "debug.leave_workspace_dir"
+	leaveWorkspaceDirectoryEnvVar    = "CSB_DEBUG_LEAVE_WORKSPACE_DIR"
 )
 
+func init() {
+	viper.BindEnv(leaveWorkspaceDirectoryConfigKey, leaveWorkspaceDirectoryEnvVar)
+	viper.SetDefault(leaveWorkspaceDirectoryConfigKey, false)
+}
+
```
zucchinidev, line 48: Hi @blgm, thank you very much for this functionality. I believe it will be a great help during debugging. I often found myself commenting out the deletion folder line when gathering more information about the executed TF code. That said, I have a few comments that I would love for you to consider.

In terms of Domain-Driven Design, using environment variables to control behavior might be seen as a violation of the principle of encapsulating behavior within domain objects. In DDD, it is generally preferable to make dependencies explicit and to encapsulate behavior within domain objects. Although I do not consider this piece of code a domain object but rather a service to assist with the infrastructure, I believe that injecting the configuration as a dependency could make the code easier to understand and test. This approach would make the dependency on the configuration explicit and would allow us to centralize the configuration logic in one place.

However, implementing this suggestion would complicate the generation of the workspace when calling the `NewWorkspace` function, which is invoked from many places. Therefore, I would suggest an intermediary solution, such as being explicit and setting the configuration in the constructor:


```go
type Workspace struct {
    // other fields...
    DebugLeaveWorkspaceDir bool
}

func NewWorkspace(...) *Workspace {
    return &Workspace{
        // initialize other fields...
        DebugLeaveWorkspaceDir: !viper.GetBool(config.leaveWorkspaceDirectoryConfigKey),
    }
}

// Then, in the function that uses this configuration:
func (w *Workspace) DoSomething() {
    // use w.DebugLeaveWorkspaceDir instead of reading from the environment
}
```

Regarding the declaration of the viper keys in this piece of code:

I would recommend declaring them in `brokerapi/broker/config.go` or near `brokerapi/broker/config.go:33`, to centralize the configuration logic. This approach would simplify maintenance by consolidating all configuration-related logic in one location, making it easier to understand and manage the application’s configuration.

zucchinidev (approved):

PR #1036, opened 2024-06-12
feat: encryption passwords as YAML in config file

blgm:
When encryption passwords are specified in a YAML config file rather than environment variables, they can now be specified as YAML rather than having to specify them as a string that is then separately parsed as JSON.


internal/encryption/passwordparser/parser.go
```diff
@@ -5,57 +5,71 @@ import (
...
+}
+
+func parsePasswordsFromMapSlice(input []any) ([]PasswordEntry, error) {
+	if len(input) == 0 {
+		return nil, nil
+	}
+
+	asJSON, err := json.Marshal(input)
+	if err != nil {
+		return nil, fmt.Errorf("error coding passowrds as JSON: %w", err)
```
zucchinidev, line 70:
```suggestion
		return nil, fmt.Errorf("error coding passwords as JSON: %w", err)
```

internal/encryption/passwordparser/parser.go
```diff
@@ -5,57 +5,71 @@ import (
...
+	var receiver []PasswordEntry
+	if err := json.Unmarshal([]byte(input), &receiver); err != nil {
+		return nil, fmt.Errorf("password configuration string could not be parsed as JSON: %w", err)
 	}
 
-	return result
+	return receiver, nil
+}
+
+func parsePasswordsFromMapSlice(input []any) ([]PasswordEntry, error) {
```
zucchinidev, line 63: praise: while there might be more efficient ways to implement this function, the current approach is a reasonable solution given its simplicity and the benefit of code reuse

zucchinidev (approved):

PR #1038, opened 2024-06-13
feat: support using actual json / yaml objects in config

nouseforaname:
allow yaml / json objects in config files for
- encrypted_passwords
- plan definitions
- provision defaults and other overrides in services

currently csb expects

`service.something.plans` to be a STRING. The code will do some manual deserializing and would fail if that path in the yaml / json config didn't contain a string.

This adds support for using real json / yaml objects to configure the csb app. 
In the current deployment process of csb via a Tile, multiple layers of config are rendered and ensuring that (e.g. plans) value stays a string is convoluted and error prone. We should just allow using actual objects instead of expecting string representations of objects..

add comments for helper functions to access viper config

[#187498188]

### Checklist:

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -17,6 +17,7 @@ package cmd
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"fmt"
 	"log/slog"
 	"net/http"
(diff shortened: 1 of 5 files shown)


blgm:
Merged via #1039

PR #1039, opened 2024-06-13
feat: allow yaml / json objects in config files for

blgm:
encrypted_passwords
plan definitions
provision defaults and other overrides in services currently csb expects

service.something.plans to be a STRING. The code will do some manual deserializing and would fail if that path in the yaml / json config didn't contain a string.

This adds support for using real json / yaml objects to configure the csb app. In the current deployment process of csb via a Tile, multiple layers of config are rendered and ensuring that (e.g. plans) value stays a string is convoluted and error prone. We should just allow using actual objects instead of expecting string representations of objects..

add comments for helper functions to access viper config

[#187498188]


blgm:
- This is a part of #1038 that I extracted
- Closes #1038
- There are some bits of this PR that look a bit strange - unmarshaling, marshaling, then unmarshaling again. I think this behavior has always been there, and this PR just calls this out. There are definitely some improvements that can be made to this code, but the tests pass, and we can improve the code later if we choose to.

PR #1041, opened 2024-06-19
fix: allow provider binaries without version in name

blgm:
Historically providers have had the version as part of the file name. But some providers don't have this. This fix allows those providers to work, but with a tradeoff that you won't be able to have more than one version of that provider.


blgm:
There's not a strong drive to deliver this right now.

Issue #1101, opened 2024-09-26
[BUG] `func NewWorkspace` does not create deep copy of `terraformTemplates`


kanngiesser:
## Description
The function `NewWorkspace` does not create a deep copy of Terraform Templates which are read from the respective service definition.

## Expected Behavior
The function should create a deep copy. 
Changes to Terraform Templates which are saved for a distinct workspace object must not affect Terraform Templates which are stored for the serviceDefinition object.

## Actual Behavior
Changes to Terraform Templates which are stored for a workspace object overwrite the serviceDefinition:

pkg/providers/tf/provision.go, line 185:
```go
workspace.Modules[0].Definitions["main"] = tf
```

## Steps to Reproduce

1.
2.
3.
4.

## Context
We created a Brokerpak which allows to provision new S3 buckets or import existing ones. The brokerpak manifest looks similar to this (excerpt):

```yml
plans:
  - name: standard
    # (...)

provision:
  import_inputs:
    - field_name: s3_bucket_name_subsume 
      type: string
      details: Existing AWS S3 bucket to subsume
      tf_resource: aws_s3_bucket.this
  import_parameters_to_delete:
    - aws_s3_bucket.this.id
    - aws_s3_bucket.this.arn
    - aws_s3_bucket.this.bucket_domain_name
    - aws_s3_bucket.this.bucket_regional_domain_name
    - aws_s3_bucket.this.hosted_zone_id
    - aws_s3_bucket.this.region
  plan_inputs:
    - field_name: subsume
      type: boolean
      details: Subsume existing
      default: false
  user_inputs: 
    - required: false
      field_name: subsume
      type: boolean
      details: "Determines whether existing bucket is imported"
      default: false
    - required: false
      field_name: s3_bucket_name_subsume
      type: string
      details: "Name of the imported bucket"
      prohibit_update: true   # bucket name cannot be changed since this would trigger a recreate
  template_refs:
    data: terraform/s3/provision/data.tf
    main: terraform/s3/provision/main.tf
    outputs: terraform/s3/provision/outputs.tf
    provider: terraform/s3/provision/provider.tf
    variables: terraform/s3/provision/variables.tf
    versions: terraform/s3/provision/versions.tf
```

The Brokerpak allows us to:
* create a new bucket with `cf create-service <service> <plan> <instance> -c '{}'`
[...]

PR #1102, opened 2024-09-27
fix: NewWorkspace creates deep copy of service definition data

kanngiesser:
`func NewWorkspace` creates deep copies of data which is read from the static service definition to prevent accidential overwrites (See https://github.com/cloudfoundry/cloud-service-broker/issues/1101)

### Checklist:


kanngiesser:
For our team this is a serious issue which blocks further rollout to production (See #1101).

Please provide some feedback for this PR, I will be happy to make changes as required.

blgm:
Hi @kanngiesser. There were some trivial test failures with this PR, but I've just fixed them in #1114 as it was very quick. As far as I know the import/subsume function is barely used, which probably explains why this was not discovered before.

PR #1103, opened 2024-09-28
feat: add validation for allowed OS and Arch combinations

kanngiesser:
Add validation for Platform Object.
- Allow only values which are valid for `$GOOS` and `$GOARCH` (required for `func CurrentPlatform()` to work)
- Allow only values for which there is an OpenTofu release (required to download binaries, see `brokerpakurl.go`) 

Update Brokerpak specification to list correct values for allowed OS and Arch

### Checklist:


blgm:
Thank you @kanngiesser

PR #1109, opened 2024-10-03
fix: removed redundant data from service_instance_details table

ifindlay-cci:
removed OperationID and OperationType from service_instance_details table.

Storing this data in the service_instance_details table is redundant, since this information is also stored the terraform_deployment table. If these two sources of truth became inconsistent, there is a possiblity that a service details could be deleted incorrectly following a failed deletion. Having a single source of truth prevents this incorrect behaviour.

[#TPCF-26508](https://vmw-jira.broadcom.com/browse/TPCF-26508)

### Checklist:

### DIFF
--- a/brokerapi/broker/deprovision.go
+++ b/brokerapi/broker/deprovision.go
@@ -94,8 +94,6 @@ func (broker *ServiceBroker) Deprovision(ctx context.Context, instanceID string,
 		return domain.DeprovisionServiceSpec{}, err
 	}
 
-	fmt.Println("serviceProvider Deprovsion")
-
 	operationID, err := serviceProvider.Deprovision(ctx, instance.GUID, vars)
 	if err != nil {
 		return domain.DeprovisionServiceSpec{}, err


brokerapi/broker/deprovision.go
```diff
@@ -94,6 +94,8 @@ func (broker *ServiceBroker) Deprovision(ctx context.Context, instanceID string,
 		return domain.DeprovisionServiceSpec{}, err
 	}
 
+	fmt.Println("serviceProvider Deprovsion")
```
blgm, line 97: Is this meant to be here?

brokerapi/broker/last_instance_operation.go
```diff
@@ -31,12 +31,11 @@ func (broker *ServiceBroker) LastOperation(ctx context.Context, instanceID strin
 
 	_, serviceProvider, err := broker.getDefinitionAndProvider(instance.ServiceGUID)
 	if err != nil {
+		broker.Logger.Info("XXXX getDefinitionAndProvider", correlation.ID(ctx), lager.Data{"err": err})
```
blgm, line 46: Is this meant to be here?

brokerapi/broker/provision.go
```diff
@@ -82,17 +83,19 @@ func (broker *ServiceBroker) Provision(ctx context.Context, instanceID string, d
 	}
 
 	// get instance details
```
blgm, line 85: I think "get instance details" doesn't make sense as a comment any more

dbservice/migrations.go
```diff
@@ -128,6 +128,14 @@ func RunMigrations(db *gorm.DB) error {
 		return autoMigrateTables(db, &models.BindRequestDetailsV1{})
 	}
 
+	migrations[16] = func() error {
```
blgm, line 131: This is doable in a single migration:
```go
migrations[16] = func() error {
  if err := db.Migrator().DropColumn(&models.ServiceInstanceDetailsV3{}, "operation_type"); err != nil {
    return err
  }
  return db.Migrator().DropColumn(&models.ServiceInstanceDetailsV3{}, "operation_id")
}
```
But I'm not such which approach is best. Given I can't find an argument that a single migration is "better" then maybe leave it as it is.

pkg/providers/tf/deprovision.go
```diff
@@ -12,6 +13,8 @@ import (
 
 // Deprovision performs a terraform destroy on the instance.
 func (provider *TerraformProvider) Deprovision(ctx context.Context, instanceGUID string, vc *varcontext.VarContext) (*string, error) {
+	fmt.Println("XXXXX YO")
```
blgm, line 16: Is this intentional?

pkg/providers/tf/deprovision.go
```diff
@@ -22,6 +25,7 @@ func (provider *TerraformProvider) Deprovision(ctx context.Context, instanceGUID
 		return nil, err
 	}
 
+	fmt.Println("provider.destror")
```
blgm, line 28: Is this intentional?

blgm (approved):
This is great. There are a few minor cosmetic improvements, but it's a good improvement that leave the code tidier - and fixes a bad bug.

brokerapi/broker/deprovision.go
ifindlay-cci: no

brokerapi/broker/last_instance_operation.go
ifindlay-cci: no

pkg/providers/tf/deprovision.go
ifindlay-cci: no

pkg/providers/tf/deprovision.go
ifindlay-cci: no

PR #1114, opened 2024-10-07
test: updates to allow #1102 to be merged

blgm:
These are some changes to tests that when merged with PR #1102 will hopefully allow the tests to continue working.


kanngiesser:
Thank you @blgm. 

I think the tests uncovered an actual problem, though: Values which are returned by the function NewWorkspace changed with #1102. Namely, the function does not return proper go zero-values in case function inputs are empty or nil.

I proposed #1115 to fix this.

PR #1128, opened 2024-10-30
chore: log to standard output

zucchinidev:
The logger, accessible through helper functions Print[f|ln], Fatal[f|ln], and Panic[f|ln], writes to standard error and includes the date and time for each logged message.

These logs are not errors, and this change directly affects the redirection of outputs and the creation of log files.

### Checklist:


internal/brokerpak/manifest/manifest.go
```diff
@@ -10,13 +10,14 @@ import (
...
-	TerraformVersions                  []TerraformVersion
-	TerraformProviders                 []TerraformProvider
+	PackVersion        int
+	Name               string
+	Version            string
+	Metadata           map[string]string
+	Platforms          []platform.Platform
+	TerraformVersions  []TerraformVersion
+	TerraformProviders []TerraformProvider
+	// TODO - Binaries array will be empty - Remove it - There are not brokerpaks with this configuration - Check internal/brokerpak/manifest/terraform_resource.go:37
```
blgm, line 20: Good spot, but this PR would be cleaner without adding a TODO.

internal/brokerpak/packer/pack.go
```diff
@@ -22,13 +21,13 @@ const manifestName = "manifest.yml"
 const binaryName = "tofu"
 
 func Pack(m *manifest.Manifest, base, dest, cachePath string, includeSource, compress bool) error {
-	// NOTE: we use "log" rather than Lager because this is used by the CLI and
+	// NOTE: we use "fmt" rather than Lager because this is used by the CLI and
```
blgm, line 24: But we're not actually using `fmt` here.

internal/brokerpak/packer/printer.go
```diff
@@ -0,0 +1,20 @@
+package packer
```
blgm, line 1: An alternative would be to do something like:
```go
var logger = log.New(os.Stdout, ..., ...)

logger.Println(...)
```
This would result in exactly the same format, but going to STDOUT.

I don't have a strong opinion, but it results in slightly less code to maintain.

pkg/brokerpak/cmd_test.go
```diff
@@ -99,21 +98,17 @@ func fakeBrokerpak() (string, error) {
 	return Pack(dir, "", true, false, platform.Platform{})
 }
 
-func ExampleValidate() {
+func TestValidate(t *testing.T) {
```
blgm, line 101: Nice change. We don't use this as an example and it makes more sense as a test.

blgm (approved):

PR #1139, opened 2024-11-18
feat: implement `service_instances` and `service_bindings` endpoints

kanngiesser:
### Checklist:

### Outline

This PR adds the endpoints [`/v2/service_instances/:instance_id`](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#fetching-a-service-instance) and [`/v2/service_instances/:instance_id/service_bindings/:binding_id`](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#fetching-a-service-binding) which are defined by the Open Service Broker API Spec and updates the Service Catalog accordingly:

#### Service Catalog

* sets the property `"instances_retrievable": true` for each service offering
* sets the property `"bindings_retrievable": true` for each offering where `"bindable"` is true
   * (with the current implementation `"bindable"` is hard-coded to be true for all service offerings, though 

#### /v2/service_instances/:instance_id

* returns offeringID, PlanID and provision parameters
* returns error codes according to OSB-API spec

#### /v2/service_instances/:instance_id/service_bindings/:binding_id

* returns only binding parameters
* returns error codes according to OSB-API spec


kanngiesser:
The `Markdown Link Check` pipeline fails because https://github.com/cloudfoundry/cloud-service-broker/search?q=toggles.Features.Toggle&type=code returns status 429 (too many requests).

Please trigger a re-run of the test

PR #1179, opened 2025-01-22
fix: allow failed upgrade to be re-tried

blgm:
Previously when an upgrade stage failed, we recorded the new (higher)
TF version in the state anyway, so if the upgrade were retried, it
would be skipped because it was already deemed to be at this version.

Now we first do an upgrade/apply at the current version to account for
potentially failed upgrades. For successful upgrades this should be a
no-op as TF is idempotent, and it's cleaner to use idempotency rather than
keep track of success/failure ourselves.

### DIFF
--- a/integrationtest/fixtures/terraform-upgrade-failure-updated/fake-provision.tf
+++ b/integrationtest/fixtures/terraform-upgrade-failure-updated/fake-provision.tf
@@ -1,0 +1,18 @@
+terraform {
+  required_providers {
+    random = {
+      source  = "registry.terraform.io/hashicorp/random"
+    }
+  }
+}
+
+variable "max" {type = number}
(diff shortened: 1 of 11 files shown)


integrationtest/fixtures/terraform-upgrade-failure-updated/fake-provision.tf
```diff
@@ -0,0 +1,18 @@
...
+      source  = "registry.terraform.io/hashicorp/random"
+    }
+  }
+}
+
+variable "max" {type = number}
+
+resource "random_integer" "priority" {
+  min = 3
+  max = var.max - 4
```
zucchinidev, line 13: praise: very intelligent

PR #1192, opened 2025-02-13
chore: use Go 1.24 tool feature

blgm:


blgm:
Superseded: https://github.com/cloudfoundry/cloud-service-broker/pull/1206

PR #1233, opened 2025-04-28
chore: separate CredHub implementation details into new package

blgm:
Details of CredHub implementation (for example, needing to create a credential and then give it permissions) had leaked into the brokerapi package. This moves the details into a new package.


ifindlay-cci (approved):
lgtm

PR #1235, opened 2025-04-29
chore: removed unused config property CH_STORE_BIND_CREDENTIALS

blgm:
The property can be read in from the environment, but it's never used in the code so has no effect.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -20,23 +20,21 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pivotal-marcela-campo (approved):

pivotal-marcela-campo:
Linter detected a dead link, but it is a false positive, as it does work. It is also not introduced by this PR. Merging.

PR #1239, opened 2025-04-30
test: CredHub CA Cert and TLS skip

blgm:
Makes sure that the `InsecureSkipTLSVerify` and `CACert` properties do what they are supposed to.

### DIFF
--- a/internal/brokerchapi/brokerchapi_test.go
+++ b/internal/brokerchapi/brokerchapi_test.go
@@ -1,10 +1,14 @@
 package brokerchapi_test
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/cloudfoundry/cloud-service-broker/v2/internal/brokerchapi"
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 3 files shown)


internal/brokerchapi/brokerchapi_test.go
```diff
@@ -229,14 +208,175 @@ var _ = Describe("Broker CredHub API", func() {
...
+		BeforeEach(func() {
+			fakeUAAServer = ghttp.NewTLSServer()
+			appendUAATokenHandler(fakeUAAServer)
+
+			fakeCredHubServer = ghttp.NewTLSServer()
+			appendStoreHandlers(fakeCredHubServer)
+		})
+
+		When("self-signed TLS on UAA", func() {
+			BeforeEach(func() {
```
pivotal-marcela-campo, line 228: this before each seems duplicated from the one in the Describe above (same for the rest of the `when` below. Should those be deleted?

internal/brokerchapi/brokerchapi_test.go
```diff
@@ -229,14 +208,175 @@ var _ = Describe("Broker CredHub API", func() {
...
+
+			It("fails with TLS validation errors", func() {
+				err := store.Save(map[string]any{"foo": "bar"}, fakeCredentialName, fakeAppGUID)
+				Expect(err).To(MatchError(ContainSubstring("tls: failed to verify certificate: x509:")))
+			})
+		})
+
+		When("self-signed TLS on CredHub", func() {
+			BeforeEach(func() {
+				fakeUAAServer = ghttp.NewServer()
```
pivotal-marcela-campo, line 244: I am not able to see how this test case is diffent from the one above.

internal/brokerchapi/brokerchapi_test.go
```diff
@@ -229,14 +208,175 @@ var _ = Describe("Broker CredHub API", func() {
...
+			BeforeEach(func() {
+				fakeUAAServer = ghttp.NewTLSServer()
+				appendUAATokenHandler(fakeUAAServer)
+
+				fakeCredHubServer = ghttp.NewTLSServer()
+				appendStoreHandlers(fakeCredHubServer)
+			})
+
+			It("fails with TLS validation errors", func() {
+				err := store.Save(map[string]any{"foo": "bar"}, fakeCredentialName, fakeAppGUID)
```
pivotal-marcela-campo, line 267: Same here, it looks like the same setup and the same assertions

pivotal-marcela-campo (changes requested):

pivotal-marcela-campo (approved):

internal/brokerchapi/brokerchapi_test.go
blgm: I've tidied it up a bit to remove a BeforeEach() that wasn't needed and highlighted how the testcases differ

PR #1248, opened 2025-05-07
chore: refactor CredHub credential store

blgm:
- Identified that we are using the repository pattern, so changes names to match
- Merges the "brokerchapi" and "brokercredstore" packages into a single layer, reducing the amount of indirection


internal/credhubrepo/credhub_repo.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	// Delete the credential
+	if err := s.http(http.MethodDelete, fmt.Sprintf("/api/v1/data?name=%s", path), nil, nil, http.StatusNoContent); err != nil {
+		return fmt.Errorf("failed to delete credential %q: %w", path, err)
+	}
+	return nil
+}
+
+func (s *Repo) http(method, path string, requestBody, responseBody any, okCodes ...int) error {
```
pivotal-marcela-campo, line 116: suggestion(non-blocking): as far as I can see okCodes is always http.StatusOK in this code. If that's the case, can we simplify and remove it?

pivotal-marcela-campo (approved):

PR #1253, opened 2025-05-16
fix: handle concurrent attempts to fetch token

blgm:
In the unlikely event that two Bind operations happen at the same time, we would need to make two calls to CredHub to store the binding, but we should only make one call to UAA to get a CredHub token.

### DIFF
--- a/internal/credhubrepo/credhub_repo.go
+++ b/internal/credhubrepo/credhub_repo.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"net/url"
 	"slices"
+	"sync"
 	"time"
 
 	"github.com/cloudfoundry/cloud-service-broker/v2/pkg/config"
@@ -20,7 +21,8 @@ type Repo struct {
 	credHubURL string
 	httpClient *http.Client
 	uaaClient  uaaClient
-	token      token
+	token      token // should only be used in loadToken() and fetchToken() that use the tokenLock mutex
+	tokenLock  sync.Mutex
 }
 
 // New creates a new CredHub repo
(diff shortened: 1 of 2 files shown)


internal/credhubrepo/credhub_repo.go
```diff
@@ -181,23 +180,51 @@ func (r *Repo) http(method, path string, requestBody, responseBody any, okCodes
 	return nil
 }
 
-func (r *Repo) updateToken() error {
+// loadToken will return an existing unexpired token if there is one, and otherwise will
+// fetch a new token (compare with fetchToken())
+//
+// Why do we use sync.Mutex and not sync.RWMutex? While a RWMutex is arguably more correct
```
pivotal-marcela-campo, line 186: praise: Thanks for this comment

internal/credhubrepo/credhub_repo_test.go
```diff
@@ -267,14 +268,21 @@ var _ = Describe("CredHub Repository", func() {
...
-				Expect(err).NotTo(HaveOccurred())
-				_, err = repo.Save(fakeCredentialPath, map[string]any{"foo": "bar"}, fakeActor)
-				Expect(err).NotTo(HaveOccurred())
+			It("it only fetches the token once", func() {
+				const calls = 5
+				var wg sync.WaitGroup
+				wg.Add(calls)
+
+				By("calling the Save() method multiple times")
+				for range calls {
```
pivotal-marcela-campo, line 277: praise: neat!

pivotal-marcela-campo (approved):

